PR #10, opened 2025-02-06
Vite + Preact + Tailwind

wszeborowskimateusz:


.github/workflows/ci.yml
```diff
@@ -7,6 +7,6 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
         with:
-          node-version: '16'
-      - run: yarn install
-      - run: yarn test
+          node-version: '18'
```
czhu12, line 10: why not just jump to v22?

src/transformers/types.ts
```diff
@@ -0,0 +1,22 @@
...
+export type ImporterTransformerDefinition =
+  | ImporterTransformerDefinitionBase
+  | CustomTransformerDefinition;
+
+export type ImporterTransformerType =
+  | 'phone_number'
+  | 'postal_code'
+  | 'state_code'
+  | 'strip'
+  | 'custom';
```
czhu12, line 12: Should dthis be an enum type to make the API easier to code against? WIth autocompletions?

src/theme/types.tsx
```diff
@@ -0,0 +1,12 @@
+export interface ImporterTheme {
```
czhu12, line 1: whats the difference between an interface and a type?

src/transformers/types.ts
wszeborowskimateusz: Well, I guess it is sort of an enum. Try it - it should autocomplete it for you + it shouldn't allow putting any other value

src/theme/types.tsx
wszeborowskimateusz: I guess mainly it is the way you define and combine them. But I think semantically they are very similar. Generally I use interfaces to define object structures and use types to define more primitive types, like enums.

src/components/Align.tsx
```diff
@@ -0,0 +1,15 @@
+import { ReactNode } from 'preact/compat';
```
czhu12: Probably don't need the align component anymore?

czhu12: lets not deal with that in this PR I guess.

src/importer/index.tsx
```diff
@@ -1,213 +1,141 @@
...
-    .map((h) => h.selectedField!.value);
+  function onMappingsSet(mappings: ColumnMapping[]) {
+    const mappedData = getMappedData(sheets, mappings, parsedFile!);
 
-  const unselectedFields = fields.filter((filter) => {
-    return usedFilters.indexOf(filter.key) === -1;
-  });
+    const newMappedData =
+      onDataColumnsMapped != null
+        ? onDataColumnsMapped(mappedData)
```
czhu12: I wonder if this should support async?

src/mapper/components/HeaderMapper.tsx
```diff
@@ -0,0 +1,83 @@
...
+        value: {
+          sheetId: sheetDefinition.id,
+          sheetColumnId: column.id,
+        },
+      };
+    });
+  });
+
+  const [mappings, setMappings] = useState<ColumnMapping[]>(
+    buildSuggestedHeaderMappings(sheetDefinitions, csvHeaders)
```
czhu12: I wonder if this should be set in the reducer, after file is uploaded, rather than in the HeaderMapper component?

czhu12: Then any updates to setMappings could dispatch back into the reducer also? I'm not sure if thats better, but it makes the reducer the place where all the business logic is stored... kinda

src/mapper/components/HeaderMapper.tsx
```diff
@@ -0,0 +1,83 @@
...
+  );
+
+  return (
+    <div>
+      <div>
+        <h5>{parsed.data.length - 1} Rows Imported</h5>
+      </div>
+      {csvHeaders.map((header, columnIndex) => {
+        const examples = data.slice(1, 4).map((d) => d[header]);
```
czhu12: Not a huge fan of this code (I know I wrote it). At the very least we should lift the 4 upwards. But also, this assumes the CSV must be a header included CSV. Also the first 3 values could be nil.

src/mapper/components/HeaderMapper.tsx
```diff
@@ -0,0 +1,83 @@
...
+}
+
+export default function HeaderMapper({
+  parsed,
+  sheetDefinitions,
+  onMappingsSet,
+}: Props) {
+  const data = parsed.data;
+  const csvHeaders = parsed.meta.fields!; // TODO THIS BRANCH: Check why it can be undefined
+  const options = sheetDefinitions.flatMap((sheetDefinition) => {
```
czhu12: Don't we have to pull out options that already belong to a mapping from this?

src/mapper/components/HeaderMapper.tsx
```diff
@@ -0,0 +1,83 @@
...
+          <Col>
+            <Button outline>Back</Button>
+          </Col>
+          <Col>
+            <Align right>
+              <Button onClick={() => onMappingsSet(mappings)}>Confirm</Button>
+            </Align>
+          </Col>
+        </Row>
+      </Margin>
```
czhu12: Bleh, can't believe I wrote it like this before. So ugly, need to refactor these into actual UI components :P

src/mapper/components/HeaderMapper.tsx
czhu12: also, `options` is a little poorly named, my eyes kinda glazed over that to mean configuration options, not selection Options

src/mapper/components/HeaderMapperSelection.tsx
```diff
@@ -0,0 +1,75 @@
+import Select from 'react-select';
```
czhu12: huh... surprised this can work without react? react must still be in the package-lock.json

czhu12: what does this mean? 
image
I guess it is? We should probably just rip out react

src/transformers/index.ts
```diff
@@ -1,50 +1,73 @@
...
+
+      newSheetStates.push({ sheetId: sheetDefinition.id, rows: newRows });
+    }
+  });
+
+  return newSheetStates;
+}
 
-export class Pipeline {
+class Pipeline {
```
czhu12, line 70: new file?

src/transformers/index.ts
```diff
@@ -1,50 +1,73 @@
...
+      row[columnId] = pipeline.transform(row[columnId]);
     });
   });
 
-  return newData;
-};
+  return sheetRows;
+}
+
+export function applyTransformations(
```
czhu12, line 47: Scared to even think about this, but i wonder if this needs to be able to be async?

src/importer/index.tsx
wszeborowskimateusz: ok, added async support

src/mapper/components/HeaderMapper.tsx
wszeborowskimateusz: Yeah, kinda, since it cannot have async operations, so things like asynchronous parsing, or asynchronous transforms cannot be in the pure reducer.

wszeborowskimateusz: Initially I was thinking that mapper could have its own reducer. But I guess I can use the global reducer for it

wszeborowskimateusz: ok, moved to global reducer

src/mapper/components/HeaderMapper.tsx
wszeborowskimateusz: Moved to constant. 

I think we need to assume that uploaded CSVs have first row as headers? Papaparse has `headers` boolean argument that we use. Depending on this argument the data is returned in different format. Do you think we need to somehow detect if uploaded file has headers row?

wszeborowskimateusz: I did add filtering out empty rows

src/mapper/components/HeaderMapper.tsx
wszeborowskimateusz: > Don't we have to pull out options that already belong to a mapping from this?

So I did it so that if you select sheel column that already has mapping, then it unselects it in the other place. I think thats how Flatfile also does it, they just have it more fancy and show selected/unselected in the dropdown 
Screenshot 2025-02-11 at 8 12 03 AM

wszeborowskimateusz: Changed the name to `mappingSelectionOptions`

src/mapper/components/HeaderMapperSelection.tsx
wszeborowskimateusz: I think vite plugin for preact adds aliases from react to preact, so when some package we use requires react as peerDependency, this should be satisfied by preact. At least thats how I understand it works

src/transformers/index.ts
wszeborowskimateusz: This is also a good question when exactly we should run those. Currently I run them on the final file upload (when user hits submit), not sure if we also want to run it for any data manipulation 🤔

wszeborowskimateusz: If we just run it at the end, then adding async shouldn't be a problem

PR #12, opened 2025-02-12
Component updates

emily-fields:


czhu12 (approved):

src/components/Button.tsx
```diff
@@ -1,56 +1,101 @@
...
+    },
+  ],
+  defaultVariants: {
+    size: 'md',
+    variant: 'primary',
+    disabled: false,
+  },
+});
+
+export default function Button2({
```
wszeborowskimateusz, line 87: Why is it called Button2 @emily-fields ?

emily-fields: It shouldn't be. I was testing it and didn't change it back https://github.com/czhu12/react-importer/pull/14

PR #13, opened 2025-02-12
Replace AgGrid with Tanstack

wszeborowskimateusz:


src/components/Tabs.tsx
```diff
@@ -0,0 +1,62 @@
...
+                key={tab.label}
+                aria-current={tab.value === activeTab ? 'page' : undefined}
+                onClick={() => onTabChange(tab.value)}
+                className={`
+                  ${
+                    tab.value === activeTab
+                      ? 'border-indigo-500 text-indigo-600'
+                      : 'border-transparent text-gray-500 hover:border-gray-300 hover:text-gray-700'
+                  }
+                  'border-b-2 px-1 py-4 text-sm font-medium whitespace-nowrap')
```
emily-fields: Do you need the `)` here? It could also be nice to add `cursor-pointer` which we have on the the button component.

wszeborowskimateusz: fixed

czhu12 (approved):

PR #15, opened 2025-02-13
Custom themes

emily-fields:


src/importer/components/FileUploader.tsx
```diff
@@ -55,15 +52,30 @@ export default function FileUploader({ setFile }: Props) {
...
                   Drag and drop your file here
                 </h5>
-                <TextStyled muted>Limit 20MB • CSV</TextStyled>
+                <div className="text-secondary">Limit 20MB • CSV</div>
               </Col>
             </Row>
           </Col>
           <Col flex="1">
             <div className="text-right">
+              <Button variant="primary">
```
czhu12: why so many buttons?

emily-fields: Oh I was just testing it. Those will be removed!

example-preact/src/types.tsx
```diff
@@ -0,0 +1,13 @@
...
+    secondary: string;
+    tertiary: string;
+    success: string;
+    danger: string;
+    warning: string;
+    info: string;
+  };
+};
+
+export type ThemeVariant = 'default' | 'theme-1' | 'theme-2';
```
wszeborowskimateusz: I think this should come from the main package?

example-preact/src/App.tsx
```diff
@@ -1,14 +1,44 @@
...
-import Importer, {
-  THEME_SIGNAL,
-  THEME_FRESCA,
-  THEME_DEFAULT,
-  ImporterTheme,
-  SheetState,
-} from 'react-importer';
+import Importer, { SheetState, ThemeVariant } from 'react-importer';
 import 'react-importer/dist/react-importer.css';
+import { ImporterTheme } from './types';
```
wszeborowskimateusz, line 4: How about example (react) app - we should also modify it the same way

example-preact/src/App.tsx
```diff
@@ -140,7 +170,7 @@ const App = () => {
             Want to see a demo? Try uploading <a href="data.csv">this file</a>.
           </h1>
           <Importer
-            theme={{ ...ALL_THEMES[currentTheme] }}
+            theme={currentTheme as ThemeVariant}
```
wszeborowskimateusz: Why do we need to cast it here? Couldn't we specify the type in useState? `useState<ThemeVariant>...`

src/importer/components/Completed.tsx
```diff
@@ -81,12 +75,10 @@ const Uploading = ({
   progress: number;
   pending?: boolean;
 }) => {
-  const theme = useTheme();
-
   return (
     <div className="text-center my-16">
       <div className="w-24 h-24 mx-auto relative">
-        <CircularProgress progress={progress} color={theme.colors.success} />
+        <CircularProgress progress={progress} color="success" />
```
wszeborowskimateusz: Yeah, I think there is no such color as `success`

src/importer/components/SubmitButton.tsx
```diff
@@ -7,13 +7,13 @@ interface Props {
 export default function SubmitButton({ onSubmit }: Props) {
   return (
     <div className="my-5">
```
wszeborowskimateusz: Do we need so many divs here?

wszeborowskimateusz: Screenshot 2025-02-14 at 7 58 49 AM

Isn't this enough? Or even just inline the button in Importer?

src/mapper/components/HeaderMapperRow.tsx
```diff
@@ -31,17 +31,17 @@ export default function HeaderMapperRow({
   return (
     <div className="my-5">
       <Card>
-        <Row>
-          <Col>
+        <div className="flex">
+          <div className="flex-1">
```
wszeborowskimateusz: Why do we need those 2 divs here?

src/mapper/components/HeaderMapperSelection.tsx
```diff
@@ -42,8 +42,8 @@ export default function HeaderMapperSelection({
                 />
               </svg>
             </div>
-          </Col>
-          <Col>
+          </div>
+          <div className="flex-1">
```
wszeborowskimateusz: Isn't there something wrong on this one? To me looks like name and dropdown aren't aligned vertically?
Screenshot 2025-02-14 at 8 02 21 AM

wszeborowskimateusz: I think Select component might be adding an unnecessary top padding?

src/theme/types.tsx
```diff
@@ -1,12 +1 @@
...
-    secondary: string;
-    success: string;
-    danger: string;
-    info: string;
-    warning: string;
-    light: string;
-    dark: string;
-  };
-}
+export type ThemeVariant = 'default' | 'theme-1' | 'theme-2';
```
wszeborowskimateusz, line 1: Do we wanna come up with better name for those? Or we want to do it later?

src/mapper/components/HeaderMapper.tsx
```diff
@@ -70,18 +70,18 @@ export default function HeaderMapper({
         );
       })}
       <div className="my-5">
-        <Row>
-          <Col>
+        <div className="flex">
```
wszeborowskimateusz: Screenshot 2025-02-14 at 8 08 29 AM


Won't that achieve the same effect with less code?

src/components/Button.tsx
```diff
@@ -17,24 +17,24 @@ interface Props {
...
+const buttonStyles = cva('text-center inline-block font-semibold', {
   variants: {
     variant: {
-      primary: 'shadow-xs bg-sky-700  text-white',
+      primary: 'shadow-xs bg-primary text-white',
       secondary:
-        'bg-white text-sky-700 ring-1 shadow-xs ring-sky-700 ring-inset',
+        'bg-white text-primary ring-1 shadow-xs ring-bg-primary ring-inset',
       tertiary:
         'bg-white text-gray-500 ring-1 shadow-xs ring-gray-300 ring-inset',
```
wszeborowskimateusz: This doesn't use tertiary colors?

src/mapper/components/HeaderMapperSelection.tsx
wszeborowskimateusz: Plus we might want to add `items-center` to the top div?

wszeborowskimateusz: Screenshot 2025-02-14 at 8 12 21 AM

Something like this (we probably don't need the inner div, so I removed it)

src/theme/types.tsx
emily-fields: I was thinking do this later especially since I think we will likely change these themes.

src/importer/components/Completed.tsx
emily-fields: There is because I set it in the custom colors. I got this part working. I made two variants but we could just have this only have the green success variant since that is all we use right now and get rid of the variant options

wszeborowskimateusz: Wait, if we always pass success then why do we have those variants in the first place?

src/theme/types.tsx
wszeborowskimateusz: ok 👍

src/importer/components/Completed.tsx
emily-fields: I was thinking that in the future we might want to have this be different colors but since right now we only have the green success variant I think it makes more sense to remove the customization

wszeborowskimateusz: YAGNI - if we need this later we will implement it - it shouldn't be difficult

wszeborowskimateusz (approved):

PR #21, opened 2025-02-14
Add an option to remove rows

wszeborowskimateusz:


src/components/ConfirmationModal.tsx
```diff
@@ -0,0 +1,85 @@
...
+                </DialogTitle>
+                {subTitle && (
+                  <div className="mt-2">
+                    <p className="text-sm text-gray-500">{subTitle}</p>
+                  </div>
+                )}
+              </div>
+            </div>
+            <div className="mt-5 sm:mt-4 sm:flex sm:flex-row-reverse">
+              <button
```
emily-fields, line 61: Seems like we should use the button component that we have for consistency. What do you think? We could do that in a future PR though

wszeborowskimateusz: Yeah + probably we would want to use the new theme variables for this and for checkbox?

emily-fields (approved):

PR #25, opened 2025-02-18
Remove example and update example-preact

emily-fields:


wszeborowskimateusz:
Why do we want to remove the example?

wszeborowskimateusz:
As discussed we should remove the react app and add to Readme how to use this package with react

wszeborowskimateusz:
Plus after we remove example app we should rename `example-preact` back to `example`

README.md
```diff
@@ -35,28 +35,108 @@ Visit [here](https://czhu12.github.io/react-importer/) to see what React Importe
 npm install --save react-importer
 ```
 
+#### React and [Vite](https://vite.dev/)
+
+- Since this project uses [Preact](https://preactjs.com/), it is necesary to add the following code to your `vite.config.ts` to ensure compatibility if your project relies on React
```
wszeborowskimateusz, line 40: Now that I think about it, I am not really sure how it works 😓 How does this package get preact/compact exactly?

wszeborowskimateusz: https://preactjs.com/guide/v10/getting-started/#aliasing-react-to-preact - make we could include this? Not all apps must use Vite

wszeborowskimateusz: I guess it somehow works via some transitive dependencies

wszeborowskimateusz: Maybe we could even include this in Readme? https://preactjs.com/guide/v10/switching-to-preact

emily-fields: Yeah, not all apps will use Vite. I'll include that link as well! I think in the future we will want to build out more detailed instructions and documentation here

wszeborowskimateusz (approved):

PR #26, opened 2025-02-18
Update styles and apply to additional components

emily-fields:


src/components/types.tsx
```diff
@@ -0,0 +1,7 @@
+export type ButtonVariant =
```
wszeborowskimateusz: I think when it's only button styles, then this type should be defined in the Button component directly

src/components/ConfirmationModal.tsx
```diff
@@ -13,16 +16,41 @@ interface Props {
...
 export default function ConfirmationModal({
   open,
   setOpen,
   title,
   subTitle,
   confirmationText = 'Confirm',
   onConfirm,
+  variant = 'default',
 }: Props) {
+  const styles: Record<
```
wszeborowskimateusz: Shouldn't this be called classes? Or baseClasses?

src/components/ConfirmationModal.tsx
```diff
@@ -13,16 +16,41 @@ interface Props {
   subTitle?: string;
   confirmationText?: string;
   onConfirm: () => void;
+  variant: 'default' | 'danger';
```
wszeborowskimateusz: `variant: VariantType;`

src/components/Button.tsx
```diff
@@ -91,8 +88,12 @@ export default function Button({
   disabled,
   onClick,
   size,
+  className,
 }: Props) {
-  const componentClassName = buttonStyles({ variant, disabled, size });
+  const componentClassName = twMerge(
+    buttonStyles({ variant, disabled, size }),
```
wszeborowskimateusz: shouldn't this be called classes or baseClasses?

wszeborowskimateusz (approved):

PR #35, opened 2025-02-19
Improve table headers and tooltips

wszeborowskimateusz:


src/sheet/components/SheetDataEditorTable.tsx
```diff
@@ -61,6 +61,17 @@ export default function SheetDataEditorTable({
     }
   }
 
+  function cellBackgroundColor(columnId: string, rowIndex: number) {
+    if (hasCellErrors(columnId, rowIndex)) {
+      return 'bg-red-100';
```
emily-fields: Can we use the custom red that is set instead? `color-danger-extra-light`

emily-fields (approved):

emily-fields (approved):

PR #36, opened 2025-02-20
Example app styling updates

emily-fields:


example/src/App.tsx
```diff
@@ -94,8 +55,8 @@ const App = () => {
             .
           </h1>
           <Importer
-            theme={currentTheme}
-            locale={currentLocale}
+            theme="default"
+            locale="en"
```
wszeborowskimateusz: Aren't those 2 default values? Do we need to pass them? From what I see theme is required, but I guess we could make it optional with default value?

example/src/components/Footer.tsx
```diff
@@ -0,0 +1,89 @@
...
+      href: '#',
+      icon: (props: any) => <BookOpenIcon {...props} />,
+    },
+    {
+      name: 'GitHub',
+      href: '#',
+      icon: (props: any) => (
+        <svg fill="currentColor" viewBox="0 0 24 24" {...props}>
+          <path
+            fillRule="evenodd"
```
wszeborowskimateusz: Maybe we could have this and youtube icons as separate components in separate files?

example/src/components/Footer.tsx
```diff
@@ -0,0 +1,89 @@
...
+              href={item.href}
+              className="text-gray-400 hover:text-gray-300"
+            >
+              <span className="sr-only">{item.name}</span>
+              <item.icon aria-hidden="true" className="size-6" />
+            </a>
+          ))}
+        </div>
+        <p className="mt-10 text-center text-sm/6 text-gray-400">
+          &copy; 2025 OneImport, Inc. All rights reserved.
```
wszeborowskimateusz: Is it legal? 😆

emily-fields: 🤷‍♀️

wszeborowskimateusz: I guess we have no rights to this name 🤷‍♂️

wszeborowskimateusz (approved):

PR #38, opened 2025-02-20
Update icons

emily-fields:
- Use HeroIcons instead of SVG icons
- Swap GitHubIcon and YouTubeIcon
- Reuse GitHubIcon

### DIFF
--- a/example/src/components/Header.tsx
+++ b/example/src/components/Header.tsx
@@ -19,7 +19,11 @@ export default function Header() {
             <TableCellsIcon className="size-6" />
             OneImport
           </div>
-          <a href="https://github.com/czhu12/react-importer" target="_blank">
+          <a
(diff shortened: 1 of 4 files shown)


example/src/components/Header.tsx
```diff
@@ -19,17 +20,7 @@ export default function Header() {
             OneImport
           </div>
           <a href="https://github.com/czhu12/react-importer" target="_blank">
```
wszeborowskimateusz: Screenshot 2025-02-20 at 12 43 32 PM

emily-fields: Hmm this error is not showing up for me

wszeborowskimateusz: Maybe you don't have some plugins installed in your IDE?

wszeborowskimateusz (approved):

PR #42, opened 2025-02-21
Update circular progress

emily-fields:
- If progress is sent then we will show the percentage of progress. If progress is not sent we will show a spinner
- Additional ESLint update

### DIFF
--- a/src/importer/components/Completed.tsx
+++ b/src/importer/components/Completed.tsx
@@ -1,9 +1,42 @@
 import { useTranslations } from '../../i18';
 import { XMarkIcon, CheckIcon } from '@heroicons/react/24/outline';
-function CircularProgress({ progress }: { progress: number }) {
+import { useState, useEffect } from 'preact/hooks';
+
+function CircularProgress({
+  progress,
+  pending,
+}: {
+  progress: number;
+  pending?: boolean;
+}) {
   const radius = 40;
   const circumference = 2 * Math.PI * radius;
   const offset = circumference - (progress / 100) * circumference;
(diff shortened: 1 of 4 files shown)


src/importer/components/Completed.tsx
```diff
@@ -1,9 +1,42 @@
...
+  progress,
+  pending,
+}: {
+  progress: number;
+  pending?: boolean;
+}) {
   const radius = 40;
   const circumference = 2 * Math.PI * radius;
   const offset = circumference - (progress / 100) * circumference;
+  const [progressUnavailable, setProgressUnavailable] = useState(false);
```
wszeborowskimateusz: Does it work with the new Try Again option on error?

src/importer/components/Completed.tsx
```diff
@@ -1,9 +1,42 @@
...
+    }
+  }, [progress]);
+
+  if (progressUnavailable) {
+    return (
+      <div className="flex justify-center">
+        <div
+          className={
+            pending
+              ? `border-success h-22 w-22 animate-spin rounded-full border-10 border-t-transparent`
```
wszeborowskimateusz: Most of those classes are the same for pending and not pending - cannot we just add if to classes that are changing?

src/importer/components/Completed.tsx
emily-fields: Yeah I merged and just tested it and it was working for me!

src/importer/components/Completed.tsx
emily-fields: I refactored this

wszeborowskimateusz (approved):

PR #47, opened 2025-02-24
Table design updates

emily-fields:


src/sheet/components/SheetDataEditor.tsx
```diff
@@ -12,14 +12,20 @@ import {
...
-import { Checkbox, ConfirmationModal } from '../../components';
+import { ConfirmationModal } from '../../components';
 import SheetDataEditorTable from './SheetDataEditorTable';
 import { TrashIcon } from '@heroicons/react/24/outline';
 import { useTranslations } from '../../i18';
 import SheetDataEditorHeader from './SheetDataEditorHeader';
 
 const columnHelper = createColumnHelper<SheetRow>();
 
+const VIEW_MODES = {
```
wszeborowskimateusz: Why don't we just use

`type VIEW_MODES = 'all' | 'errors' | 'valid'`?

wszeborowskimateusz: Plus it could be defined in `sheet/types`?

src/sheet/components/SheetDataEditor.tsx
```diff
@@ -98,23 +112,55 @@ export default function SheetDataEditor({
...
-            id={`Only show errors checkbox for ${sheetDefinition.id}`}
-            checked={onlyShowErrors}
-            setChecked={(checked) => {
-              setSelectedRows([]);
-              setOnlyShowErrors(checked);
-            }}
-            label={t('sheet.onlyShowErrorsCheckboxLabel')}
-          />
+          <span className="isolate inline-flex rounded-md shadow-xs">
+            <button
```
wszeborowskimateusz: Maybe we could extract this button into some component? Most of the classes are the same for all of them?

src/sheet/components/SheetDataEditorTable.tsx
```diff
@@ -59,16 +59,16 @@ export default function SheetDataEditorTable({
...
   const headerClass =
-    'sticky top-0 bg-white py-3.5 pr-3 pl-4 text-left text-sm font-semibold text-gray-900 z-10';
+    'sticky top-0 bg-gray-100 py-3.5 pr-3 pl-4 text-left text-sm font-semibold text-gray-900 z-10 whitespace-nowrap ';
   const cellClass = 'text-sm font-medium whitespace-nowrap text-gray-900';
 
   return (
-    <div className="max-h-[80vh] overflow-x-auto">
-      <table className="min-w-full divide-y divide-gray-300">
-        <thead>
+    <div className="max-h-[80vh] overflow-x-auto overflow-y-auto">
```
wszeborowskimateusz: Isn't it just overflow-auto (to combine both x and y)?

src/sheet/components/SheetDataEditor.tsx
```diff
@@ -98,23 +112,55 @@ export default function SheetDataEditor({
...
+              type="button"
+              className={`relative inline-flex cursor-pointer items-center rounded-l-md bg-white px-3 py-2 text-sm font-semibold text-gray-900 ring-1 ring-gray-300 ring-inset hover:bg-gray-50 focus:z-10 ${
+                viewMode === VIEW_MODES.ALL ? '!bg-gray-900 text-white' : ''
+              }`}
+              onClick={() => {
+                setSelectedRows([]);
+                setViewMode(VIEW_MODES.ALL);
+              }}
+            >
+              All
```
wszeborowskimateusz: Those texts should come from translations

src/sheet/types.ts
```diff
@@ -52,3 +52,9 @@ export interface SheetState {
   sheetId: string;
   rows: SheetRow[]; // key being column id
 }
+
+export enum SheetViewMode {
```
wszeborowskimateusz: Hmm, we don;t use enum in other places - like e.g. look how ButtonVariant is defined

wszeborowskimateusz: I think we could keep in consistent

wszeborowskimateusz: https://stackoverflow.com/questions/40275832/typescript-has-unions-so-are-enums-redundant

src/sheet/types.ts
```diff
@@ -52,3 +52,5 @@ export interface SheetState {
   sheetId: string;
   rows: SheetRow[]; // key being column id
 }
+
+export type SheetViewMode = 'all' | 'valid' | 'errors';
```
emily-fields, line 56: @wszeborowskimateusz is this how you would prefer that this be done?

wszeborowskimateusz: Yeah

src/sheet/components/SheetDataEditor.tsx
wszeborowskimateusz: @emily-fields not sure if you missed it?

emily-fields: I can refactor this

wszeborowskimateusz: Yeah, or at least extract the common classes, but I think extreacting the whole componennt - like TabSwitcher or something like this

wszeborowskimateusz (approved):

PR #49, opened 2025-02-25
Group SelectDropdown Options by Used and Unused in Mapper

emily-fields:


src/mapper/components/HeaderMapperRow.tsx
```diff
@@ -6,22 +6,26 @@ import HeaderMapperSelection from './HeaderMapperSelection';
 interface Props {
   csvHeader: string;
   mappingSelectionOptions: MapperOption[];
+  usedMappingSelectionOptions: MapperOption[];
```
czhu12: I wonder if instead of passing two lists of `MapperOption`, we can just pass 1 in and filter it in the actual HeaderMapper?

src/mapper/utils.ts
```diff
@@ -117,24 +117,16 @@ export function getMappingAvailableSelectOptions(
           sheetId: sheetDefinition.id,
           sheetColumnId: column.id,
         },
+        used:
+          currentMapping.some(
+            (mapping) =>
+              mapping.sheetId === sheetDefinition.id &&
+              mapping.sheetColumnId === column.id
+          ) ?? false,
```
wszeborowskimateusz: when does `?? false` part is used? Can `some` function ever return null or undefined?

src/components/Select.tsx
```diff
@@ -59,14 +63,29 @@ export default function Select<T>({
 
   const selectedOptions = options.filter((option) => isSelected(option.value));
 
+  const hasUsedProperty = options.some((option) => option.used !== undefined);
+
+  const groupedOptions = hasUsedProperty
+    ? [
```
wszeborowskimateusz: I don't think it is the role of Select component to have used/unused definitions. I think Select component should rather just allow grouping options and it should be a component that uses the Select to define that those groupings are used/unused

src/components/Select.tsx
```diff
@@ -59,14 +63,29 @@ export default function Select<T>({
 
   const selectedOptions = options.filter((option) => isSelected(option.value));
 
+  const hasUsedProperty = options.some((option) => option.used !== undefined);
```
wszeborowskimateusz: What if someone passes used:false, then hasUsedProperty would be true, right? I think we should check `option.used === true` instead?

src/mapper/utils.ts
```diff
@@ -105,8 +105,10 @@ export function calculateMappingExamples(
     .slice(0, NUMBER_OF_EXAMPLES_IN_MAPPING);
 }
 
-export function getMappingAvailableSelectOptions(
-  sheetDefinitions: SheetDefinition[]
+export function useMappingAvailableSelectOptions(
+  sheetDefinitions: SheetDefinition[],
+  currentMapping: ColumnMapping[],
+  t: TranslationFunction
```
wszeborowskimateusz: Why do we need to pass it? Cannot we simply call useTranslations here?

src/i18/types.ts
```diff
@@ -9,3 +9,8 @@ export type Resources = {
 export type TranslationContextType = {
   t: (key: string, argumentValues?: ArgumentsType) => string;
 };
+
+export type TranslationFunction = (
```
wszeborowskimateusz: We will probably not need it when we call useTranslations directly

wszeborowskimateusz (approved):

wszeborowskimateusz (approved):

PR #50, opened 2025-02-25
Create new Tooltip component

emily-fields:
- Creates a new Tooltip component and renames the current Tooltip to SheetTooltip

### DIFF
--- a/src/sheet/components/SheetDataEditor.tsx
+++ b/src/sheet/components/SheetDataEditor.tsx
@@ -16,6 +16,7 @@ import {
   ConfirmationModal,
   ButtonGroup,
   ButtonGroupType,
+  Tooltip,
 } from '../../components';
 import SheetDataEditorTable from './SheetDataEditorTable';
 import {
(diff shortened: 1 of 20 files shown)


src/sheet/components/SheetDataEditor.tsx
```diff
@@ -149,17 +150,21 @@ export default function SheetDataEditor({
...
         </div>
 
-        {/* TODO: Add tooltip when disabled */}
-        <TrashIcon
-          className={`ml-8 h-6 w-6 ${
-            selectedRows.length > 0 ? 'cursor-pointer' : disabledButtonClasses
-          }`}
-          onClick={() => setRemoveConfirmationModalOpen(true)}
-        />
+        <div className="group relative">
```
wszeborowskimateusz: Shouldn't this be inside the tooltip component and we would pass TrashIcon as a child of it?

src/sheet/components/SheetDataEditor.tsx
```diff
@@ -149,17 +150,21 @@ export default function SheetDataEditor({
...
+          <TrashIcon
```
wszeborowskimateusz: Isn't the horizontal padding now different than for other action icons?

src/sheet/components/SheetDataEditor.tsx
```diff
@@ -149,17 +150,21 @@ export default function SheetDataEditor({
...
+            className={`m-2 h-6 w-6 ${
+              selectedRows.length > 0 ? 'cursor-pointer' : disabledButtonClasses
+            }`}
+            onClick={() => setRemoveConfirmationModalOpen(true)}
+          />
+          {selectedRows.length === 0 && (
```
wszeborowskimateusz: Shouldn't we add similar tooltip for download icon?

src/sheet/components/SheetDataEditor.tsx
```diff
@@ -149,17 +150,21 @@ export default function SheetDataEditor({
...
+            <Tooltip tooltipText={t('sheet.removeRowsTooltip')} />
```
wszeborowskimateusz: Do we also want to add a tooltip to disabled confirm button on the mapper page? Or we gonna handle it differently there?

emily-fields: I added this. Maybe we will want to do this differently in the future? 🤷‍♀️

src/sheet/components/SheetDataEditorActions.tsx
```diff
@@ -137,22 +138,31 @@ export default function SheetDataEditorActions({
...
-      <ArrowDownTrayIcon
-        className={`mx-5 h-6 w-6 ${
-          rowData.length > 0 ? 'cursor-pointer' : disabledButtonClasses
-        }`}
-        onClick={() => downloadSheetAsCsv(sheetDefinition, rowData)}
-      />
+      <Tooltip
+        className="ml-5"
+        tooltipText={t('sheet.removeRowsTooltip')}
+        hidden={selectedRows.length !== 0}
```
wszeborowskimateusz: Maybe `<= 0` ?

wszeborowskimateusz (approved):

PR #53, opened 2025-02-25
Filter by error

wszeborowskimateusz:


src/utils/index.ts
```diff
@@ -13,3 +13,7 @@ export const filterEmptyRows = (state: SheetState) => {
 export const isEmptyCell = (value: any) => {
   return isUndefinedOrNull(value) || value === '';
 };
+
+export const removeDupplicates = (array: any[]) => {
```
emily-fields: I think this should be `removeDuplicates` instead

wszeborowskimateusz: fixed

emily-fields (approved):

PR #57, opened 2025-02-26
Mapper page updates

emily-fields:


src/components/Badge.tsx
```diff
@@ -0,0 +1,25 @@
+import { cva } from 'cva';
+import { ReactNode } from 'preact/compat';
```
wszeborowskimateusz, line 2: Can we add an empty line below imports?

src/mapper/utils.ts
```diff
@@ -96,12 +99,41 @@ export function calculateMappingExamples(
...
+
+  const paddedExamples = padExamples(examples);
+
+  return trimExamplesByCharacterLimit(paddedExamples);
+}
+
+function getFilteredExamples(data: CSVParsedData[], csvColumnName: string) {
   return data
     .map((d) => d[csvColumnName])
     .filter((v) => v != null && v.trim() !== '')
```
wszeborowskimateusz, line 112: You might need to merge master to your branch? I think this code got updated?

src/mapper/utils.ts
```diff
@@ -96,12 +99,41 @@ export function calculateMappingExamples(
...
     .slice(0, NUMBER_OF_EXAMPLES_IN_MAPPING);
 }
 
+function padExamples(examples: string[]) {
+  return [
+    ...examples,
+    ...Array(NUMBER_OF_EXAMPLES_IN_MAPPING - examples.length).fill(''),
```
wszeborowskimateusz, line 119: Does it work with negative values?

src/mapper/utils.ts
```diff
@@ -96,12 +99,41 @@ export function calculateMappingExamples(
...
+}
+
+function trimExamplesByCharacterLimit(examples: string[]) {
+  let totalCharacters = examples.reduce((acc, curr) => acc + curr.length, 0);
+
+  while (
+    totalCharacters > MAX_CHARACTERS_IN_MAPPING_EXAMPLES &&
+    examples.length > 1
+  ) {
+    examples.pop();
```
wszeborowskimateusz: I don't think its good to modify the passed array? Shouldn't you make a local copy of it?

src/mapper/components/HeaderMapper.tsx
```diff
@@ -27,6 +30,7 @@ export default function HeaderMapper({
   onBack,
 }: Props) {
   const { t } = useTranslations();
+  const [hoveredHeader, setHoveredHeader] = useState<string | null>(null);
```
wszeborowskimateusz: maybe hoveredCsvHeader to make it consistent with the rest of the app?

src/mapper/utils.ts
emily-fields: What do you mean exactly?

src/mapper/utils.ts
emily-fields: Merged this and double checked to make sure it is working still

src/mapper/utils.ts
emily-fields: I refactored this if you could take another look

src/mapper/utils.ts
wszeborowskimateusz: Like, what happens when you pass e.g. `Array(-2).fill('')`

emily-fields: How could this happen?

wszeborowskimateusz: ok, maybe it cannot 🤷

wszeborowskimateusz (approved):

src/mapper/utils.ts
emily-fields: Do you think this is okay?

wszeborowskimateusz: Screenshot 2025-02-27 at 9 31 58 AM

wszeborowskimateusz: If you sure it can never be negative, then I think it is ok

PR #60, opened 2025-02-27
Fx scroll

emily-fields:


src/importer/index.tsx
```diff
@@ -45,6 +48,17 @@ function ImporterBody({
     dispatch,
   ] = useReducer(reducer, buildInitialState(sheets));
 
+  useEffect(() => {
+    if (targetRef.current) {
+      if (
+        mode === 'preview' &&
+        sheetData.some((sheet) => sheet.rows.length > 0)
+      ) {
+        targetRef.current.scrollIntoView({ behavior: 'smooth' });
```
wszeborowskimateusz: Looks good, maybe we could just add safety for null? `target.current?.scrollIntoView...`

wszeborowskimateusz (approved):

src/importer/index.tsx
emily-fields: Okay, I added that!

PR #63, opened 2025-02-28
Scroll to top

emily-fields:


src/importer/index.tsx
```diff
@@ -50,8 +50,8 @@ function ImporterBody({
 
   useEffect(() => {
     if (
-      mode === 'preview' &&
-      sheetData.some((sheet) => sheet.rows.length > 0)
+      mode !== 'preview' ||
+      (mode === 'preview' && sheetData.some((sheet) => sheet.rows.length > 0))
     ) {
       targetRef.current?.scrollIntoView({ behavior: 'smooth' });
```
wszeborowskimateusz: How does it work when we edit some data in the preview mode?

wszeborowskimateusz: Is it called then? (I assume sheetData changes?)

emily-fields: Yeah, nice catch! Looks like we missed that yesterday. What do you think of this?

wszeborowskimateusz: Why do we need the check for sheetData rows that are not empty? What is the use case here?

emily-fields: Since the empty table and the table with data share the same mode if we do not do this then the page will scroll to the top when we load it

wszeborowskimateusz: So basically the actual logic we want is to run on any mode change, but don't run in the initial render?
Screenshot 2025-02-28 at 11 51 29 AM

wszeborowskimateusz (approved):

PR #64, opened 2025-02-28
Fix select mapper rendering

emily-fields:


src/sheet/components/SheetDataEditorActions.tsx
```diff
@@ -168,7 +167,7 @@ export default function SheetDataEditorActions({
         clearable
         displayPlaceholderWhenSelected
         placeholder={t('sheet.filterByError')}
-        classes="min-w-48"
```
wszeborowskimateusz: Are you sure thats the correct fix for it? What if in some language the translation is longer? Cannot we try position the icon in a row somehow instead of using position absolute?

src/components/Select.tsx
```diff
@@ -126,14 +126,17 @@ export default function Select<T>({
...
         {!searchable && (
-          <ComboboxButton
-            className={`${classes} ${clearButtonDisplayed ? 'pr-8' : 'pr-2'} focus:outline-primary grid w-full cursor-default grid-cols-1 rounded-md bg-white py-1.5 pl-3 text-left text-gray-900 outline-1 -outline-offset-1 outline-gray-300 focus:outline-2 focus:-outline-offset-2 sm:text-sm/6`}
-          >
-            <span className="col-start-1 row-start-1 truncate pr-6">
-              {selectedOptions.length > 0
-                ? `${displayPlaceholderWhenSelected ? `${placeholderValue}: ` : ''}${displayValue}`
-                : placeholderValue}
-            </span>
+          <ComboboxButton className="w-full">
```
wszeborowskimateusz: Isn't that looks very similar to searchable option? Cannot we then reuse it and the only pass props that actually changes from searchable and non searchable?

wszeborowskimateusz: I assume it would be 
- displayValue
- onChange
- readOnly

src/components/Select.tsx
```diff
@@ -112,38 +125,25 @@ export default function Select<T>({
...
-              {selectedOptions.length > 0
-                ? `${displayPlaceholderWhenSelected ? `${placeholderValue}: ` : ''}${displayValue}`
-                : placeholderValue}
-            </span>
-          </ComboboxButton>
-        )}
+        <ComboboxButton className="w-full">
+          <ComboboxInput
+            className={`${classes} focus:outline-primary block w-full cursor-default truncate rounded-md bg-white py-1.5 ${clearButtonDisplayed ? 'pr-12' : 'pr-2'} pl-3 text-left text-gray-900 outline-1 -outline-offset-1 outline-gray-300 focus:outline-2 focus:-outline-offset-2 sm:text-sm`}
+            displayValue={() => getDisplayValue()}
```
wszeborowskimateusz: Cannot we just pass displayValue={getDisplayValue}?

src/sheet/components/SheetDataEditorActions.tsx
wszeborowskimateusz: This was fixed, right @emily-fields ?

wszeborowskimateusz (approved):

PR #65, opened 2025-03-02
remove header when in upload mode

czhu12:


src/importer/index.tsx
```diff
@@ -155,6 +156,7 @@ function ImporterBody({
     dispatch({ type: 'MAPPING' });
   }
 
+  console.log(mode);
```
wszeborowskimateusz: This should probably be removed

src/importer/index.tsx
```diff
@@ -168,6 +170,21 @@ function ImporterBody({
             onBack={onBackToPreview}
           />
         )}
+        {mode === 'upload' && (
+          <div className="mt-5">
+            <FileUploader setFile={onFileUploaded} />
+            {allowManualDataEntry && (
```
wszeborowskimateusz, line 175: Does this setting also need/should say if user can manually edit data after mapping (in preview mode)?

What is in general the purpose of introducing this flag?

czhu12: Yeah I think I need to disable this for lab connect for now -- yeah w eshould still allow data editing but i doubt anyone will want to manually enter data like this

wszeborowskimateusz (approved):

PR #75, opened 2025-03-05
Revert "Allow react nodes in translations"

czhu12:
Reverts czhu12/react-importer#71


wszeborowskimateusz:
I think I am gonna close it, as we would just fix the issue on master

PR #78, opened 2025-03-06
nobuild updates

czhu12:


vite.config.ts
```diff
@@ -8,6 +8,15 @@ import tailwindcss from '@tailwindcss/vite';
 // https://vite.dev/config/
 export default defineConfig({
   plugins: [tailwindcss(), preact(), dts()],
+  define: {
```
wszeborowskimateusz, line 11: What is this one use for?

wszeborowskimateusz:
I think we can close this one since we created https://github.com/czhu12/react-importer/pull/83

PR #80, opened 2025-03-06
Scope Tailwind

wszeborowskimateusz:


emily-fields:
Small nitpick is that I think we should update the README with these new variables. Otherwise it looks good to me!

wszeborowskimateusz:
> Small nitpick is that I think we should update the README with these new variables. Otherwise it looks good to me!

Are you sure about that? I don't think they changed? I think we only updated our internal variables?

emily-fields:
> > Small nitpick is that I think we should update the README with these new variables. Otherwise it looks good to me!
> 
> Are you sure about that? I don't think they changed? I think we only updated our internal variables?

You are right. I just tested it

emily-fields (approved):

PR #87, opened 2025-03-13
Fix GH Pages example file download

wszeborowskimateusz:


example/src/components/importers/WideImporter.tsx
```diff
@@ -4,6 +4,7 @@ import Importer, {
   SheetRow,
   ImporterState,
 } from 'react-importer/peer';
+import studnentsAndSchoolsExample from '../../assets/datasets/students_and_schools.csv?url';
```
emily-fields: I think we should change this to `studentsAndSchoolsExample`

wszeborowskimateusz: ohh, you mean the typo?

wszeborowskimateusz: fixed

emily-fields (approved):

PR #94, opened 2025-03-16
Uploader page

emily-fields:


src/i18/en/translation.json
```diff
@@ -33,6 +33,14 @@
       "cancelText": "No, stay here"
     }
   },
+  "uploader": {
```
wszeborowskimateusz, line 30: Shoudln't we move importer.uploader translations to this namespace?

src/sheet/types.ts
```diff
@@ -25,6 +25,7 @@ interface SheetColumnBaseDefinition {
   isReadOnly?: boolean;
   validators?: ImporterValidatorDefinition[];
   transformers?: ImporterTransformerDefinition[];
+  importerInformation?: string;
```
wszeborowskimateusz: I think we should add it to the Readme documentation?

src/uploader/types.ts
```diff
@@ -0,0 +1,11 @@
+export type ImporterRequirementType = {
```
wszeborowskimateusz, line 1: Shouldn't we export those types in src/types/index.ts and use that file as barrel file for all type imports?

src/uploader/utils.ts
```diff
@@ -0,0 +1,34 @@
...
+    sheet.columns.forEach((column) => {
+      const requirement = {
+        sheetId: sheet.id,
+        columnId: column.id,
+        columnLabel: column.label,
+        importerInformation: column.importerInformation,
+      };
+
+      if (
+        column.validators?.some(
```
wszeborowskimateusz: I think we do have an util `validators/index.ts::fieldIsRequired`

src/uploader/utils.ts
```diff
@@ -0,0 +1,34 @@
...
+          (validator) => validator.validate === 'required'
+        )
+      ) {
+        groups.required.push(requirement);
+      } else if (column.suggestedMappingKeywords) {
```
wszeborowskimateusz: Why do we only push columns with suggestedMappingKeywords as optional?

src/uploader/types.ts
```diff
@@ -0,0 +1,11 @@
...
+  sheetId: string;
```
wszeborowskimateusz, line 2: Is this field used anywhere?

src/importer/types.ts
```diff
@@ -24,6 +24,7 @@ export interface ImporterDefinition {
     onProgress: (progress: number) => void
   ) => Promise<void>;
   locale?: string;
+  importerInformation?: string;
```
wszeborowskimateusz: I think we should also add it to README documentation

src/index.css
```diff
@@ -29,6 +30,8 @@
     --color-csv-importer-danger-extra-light: var(--csv-importer-color-danger-extra-light);
     --color-csv-importer-warning: var(--csv-importer-color-warning);
     --color-csv-importer-info: var(--csv-importer-color-info);
+    --color-csv-importer-info-foreground: var(--csv-importer-color-info-foreground);
+    --color-csv-importer-muted: var(--csv-importer-color-muted);
```
wszeborowskimateusz: Shouldn't we add those to README documentation?

src/components/Tooltip.tsx
```diff
@@ -9,31 +10,83 @@ interface Props {
...
-        {tooltipText}
-        <div className="absolute top-0 left-1/2 h-2 w-2 -translate-x-1/2 -translate-y-1 rotate-45 bg-gray-900"></div>
-      </div>
+      {tooltipContainer &&
+        createPortal(
+          <div
+            ref={tooltipRef}
+            className={baseClasses({ visible: isVisible, hidden })}
+            style={{
+              position: 'absolute',
```
wszeborowskimateusz: Isn;t this added by baseClasses?

src/components/Tooltip.tsx
```diff
@@ -9,31 +10,83 @@ interface Props {
...
+              top: `${position.top}px`,
+              left: `${position.left}px`,
```
wszeborowskimateusz, line 80: Does it mean that tooltip would always be displayed below the element? What if we are e.g at the very end of the page?

src/components/Card.tsx
```diff
@@ -1,17 +1,38 @@
+import { cva } from 'cva';
 import { forwardRef } from 'preact/compat';
-import { JSX } from 'preact';
+import { ReactNode } from 'preact/compat';
```
wszeborowskimateusz: I think we can combine those 2 imports? Isn't is highlighted by your linter?
Screenshot 2025-03-17 at 9 50 39 AM

src/components/Card.tsx
```diff
@@ -1,17 +1,38 @@
...
+      },
+    },
+    defaultVariants: {
+      variant: 'default',
+    },
+  }
+);
+
+const Card = forwardRef<HTMLDivElement, Props>(
+  ({ children, className, variant, ...props }, ref) => {
```
wszeborowskimateusz: Do we use the `...props` part anywhere? I think if we don't we could get rid of it?

src/components/Card.tsx
```diff
@@ -1,17 +1,38 @@
...
+  variant?: 'default' | 'gray';
+}
+
+const baseClasses = cva(
+  'overflow-hidden rounded-md border border-gray-200 px-4 py-5 sm:p-6',
+  {
+    variants: {
+      variant: {
+        default: 'bg-white',
+        gray: 'bg-csv-importer-muted',
```
wszeborowskimateusz: Maybe the variant should be called `muted`?

src/uploader/components/ImporterRequirements.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { Tooltip } from '../../components/index';
+import Alert from '../../components/Alert';
+import { useTranslations } from '../../i18';
+
+interface Props {
+  importerRequirements: ImporterRequirementsType;
+  importerInformation?: string;
+}
+
+const RequirementInfo = ({
```
wszeborowskimateusz: Shouldn't this be a separate file?

src/uploader/components/ImporterRequirements.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+  return (
+    <div className="h-full w-full space-y-5 overflow-y-auto">
+      {Object.entries(importerRequirements).map(([groupName, requirements]) => (
+        <div key={groupName} className="me-3">
+          <div className="my-3 border-b border-gray-200 pb-4 text-sm font-light uppercase">
+            {t(`uploader.${groupName}Columns`)}
+          </div>
+          <div className="mt-4">
+            {requirements.map((requirement) => (
```
wszeborowskimateusz: What if some group is empty? do we want to display the title for it?

src/components/Card.tsx
emily-fields: Yeah just hadn't cleaned this up yet!

src/components/Card.tsx
emily-fields: Removed

src/uploader/types.ts
emily-fields: No, I'll remove

src/uploader/types.ts
emily-fields: Updated

src/uploader/utils.ts
emily-fields: We don't! I hadn't gotten around to double checking things yet. You were too fast 😆

src/uploader/components/ImporterRequirements.tsx
emily-fields: Hadn't gotten around to this yet when you reviewed it but should be fixed now!

src/components/Tooltip.tsx
emily-fields: Which part are you talking about?

src/components/Tooltip.tsx
emily-fields: Yeah right now this is the only form we support. In the future, I think we should add others but for this one I focused on fixing the issue where the toolTip was being cut off

emily-fields: If you want to make it so this displays above if it is going to be at the bottom feel free to push a change to this branch otherwise I was thinking we could tackle this in a future PR?

src/components/Tooltip.tsx
wszeborowskimateusz: position: absolute. I can already see `absolute` class in the baseClasses. Or am I missing something?

src/uploader/components/FileUploader.tsx
```diff
@@ -0,0 +1,86 @@
...
+          onDragOver={(e) => e.preventDefault()}
+          onDrop={handleDrop}
+          className="flex h-full flex-col"
+        >
+          <div className="flex flex-1 flex-col items-center justify-center">
+            <CloudArrowUpIcon className="text-csv-importer-primary mb-3 h-12 w-12" />
+            <p className="mb-3">{t('uploader.dragAndDrop')}</p>
+            <div className="text-sm text-gray-500">
+              {tHtml('uploader.fileSizeLimit', {
+                limit: <b>{'20MB'}</b>,
```
wszeborowskimateusz: Maybe we could extract the 20MB to constants?

src/uploader/components/ImporterRequirements.tsx
```diff
@@ -0,0 +1,28 @@
+import { ImporterRequirementsType } from '../../types';
+import { Alert } from '../../components';
+import { useTranslations } from '../../i18';
+import RequirementsList from './RequirementsList';
```
wszeborowskimateusz, line 4: Can we add an empty line between imports and Props?

src/uploader/components/RequirementsList.tsx
```diff
@@ -0,0 +1,46 @@
+import { ImporterRequirementsType } from '../../types';
+import { InformationCircleIcon } from '@heroicons/react/24/outline';
+import { Tooltip } from '../../components/index';
+import { useTranslations } from '../../i18';
+
+export default function RequirementsList({
+  importerRequirements,
+}: {
+  importerRequirements: ImporterRequirementsType;
```
wszeborowskimateusz, line 7: For consistency - could we extract it into Props interface?

src/uploader/components/RequirementsList.tsx
```diff
@@ -0,0 +1,46 @@
...
+        .filter(([, requirements]) => requirements.length > 0)
+        .map(([groupName, requirements]) => (
+          <div key={groupName} className="me-3">
+            <div className="my-3 border-b border-gray-200 pb-4 text-sm font-light uppercase">
+              {t(`uploader.${groupName}Columns`)}
+            </div>
+            <div className="mt-4">
+              {requirements.map((requirement) => (
+                <div
+                  key={requirement.columnId}
```
wszeborowskimateusz: Two questions here
- are we sure columnId is unique across different sheets?
- are we sure that columnLabel would be unique across different sheets? I can imagine having e.g. `id` columns in two different sheets. Maybe we could introduce something like mappingLabel and then use `mappingLabel ?? label` in mapping and this page and use `label` for table page?

src/uploader/components/FileUploader.tsx
emily-fields: Is this true? Do we actually impose this limit in any way?

wszeborowskimateusz: I don't think we do

src/components/Tooltip.tsx
emily-fields: Tested then removed this one

src/uploader/components/FileUploader.tsx
emily-fields: Maybe we should remove it since we don't enforce it at all?

wszeborowskimateusz: Yeah, I guess we can remove it

README.md
```diff
@@ -150,6 +150,7 @@ import 'react-importer/peer/index.css';
 | allowManualDataEntry            | false         | boolean                                                                          | No       | Whether to allow users to manually enter data during the preview phase                                                                     |
 | locale                          | en            | 'en' \| 'fr'                                                                     | No       | Locale string for internationalization                                                                                                     |
 | preventUploadOnValidationErrors | false         | boolean \| `(errors: ImporterValidationError[]) => boolean`                      | No       | Controls whether to prevent upload when validation errors occur. Can be a boolean or a function that returns a boolean based on the errors |
+| maxFileSize                     | -             | number                                                                           | No       | Maximum allowed file size in bytes (optional). When specified, files larger than this size will be rejected. Example: 204800 for 200KB     |
```
wszeborowskimateusz: Maybe istead of explaining in the readme that it is in bytes, just call the variable `maxFileSizeInBytes`?

wszeborowskimateusz: Shouldn't we have some default?

src/uploader/utils.ts
```diff
@@ -0,0 +1,52 @@
...
+      } else {
+        groups.optional.push(requirement);
+      }
+    });
+  });
+
+  return groups;
+}
+
+export const formatFileSize = (bytes: number): string => {
```
wszeborowskimateusz, line 32: Cannot bytes be nil currently? I am in for having the default and remove `?` from any internal use

README.md
emily-fields: Should we do 200KB?

emily-fields: I am happy with a default!

wszeborowskimateusz: Shouldn't we have our previous 20MB? Actually I am not sure why it was 20MB to begin with, but we could ask Chris. Maybe he had some reason for it?

README.md
```diff
@@ -150,6 +150,7 @@ import 'react-importer/peer/index.css';
 | allowManualDataEntry            | false         | boolean                                                                          | No       | Whether to allow users to manually enter data during the preview phase                                                                     |
 | locale                          | en            | 'en' \| 'fr'                                                                     | No       | Locale string for internationalization                                                                                                     |
 | preventUploadOnValidationErrors | false         | boolean \| `(errors: ImporterValidationError[]) => boolean`                      | No       | Controls whether to prevent upload when validation errors occur. Can be a boolean or a function that returns a boolean based on the errors |
+| maxFileSizeInBytes              | 20971520      | number                                                                           | No       | Maximum allowed file size in bytes (optional). When specified, files larger than this size will be rejected. Example: 204800 for 200KB     |
```
wszeborowskimateusz: In the default we could write is parenthesis what is the MB value?

README.md
emily-fields: Okay after discussing with Chris let's stick with 20MB for now

src/uploader/utils.ts
emily-fields: Updated

README.md
emily-fields: Do you think I should still remove the explanation that it is in bytes?

example/src/components/importers/EmployeeImporter.tsx
```diff
@@ -29,6 +29,7 @@ export default function EmployeeImporter() {
       </h1>
       <div className="flex max-h-[800px]">
         <Importer
+          maxFileSizeInBytes={10485760} // 10mb
```
wszeborowskimateusz: I think if you specify it as `10 * 1024 * 1024` it might be more readable?

src/importer/index.tsx
```diff
@@ -32,6 +32,7 @@ function ImporterBody({
   sheets,
   onDataColumnsMapped,
   preventUploadOnValidationErrors,
+  maxFileSizeInBytes = 20971520, // 20mb
```
wszeborowskimateusz: 20 * 1024 * 1024 ?

src/uploader/utils.ts
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  return `${Math.round(size)} ${units[unitIndex]}`;
+};
+
+export const isFileBelowMaxSize = (
+  fileSize: number,
+  maxSize: number
+): boolean => {
+  return fileSize <= maxSize;
```
wszeborowskimateusz: Do we need this function now when maxSize cannot be nil?

README.md
wszeborowskimateusz: no, I think explanation is fine

src/uploader/utils.ts
emily-fields: I removed it to simplify things

src/importer/index.tsx
emily-fields: I double checked it on this chart https://www.gbmb.org/mb-to-bytes 
image

wszeborowskimateusz: ?

wszeborowskimateusz: I mean, instead of writing `20971520`, you could write `20 * 1024 * 1024`, which should be the same value, but it should be more readable IMO

emily-fields: Oh I see. I'll make that change. Do you think the README should have that as well or no?

wszeborowskimateusz: I think in you do include `(20MB)`, don't you?

wszeborowskimateusz (approved):

PR #97, opened 2025-03-17
Create .nvmrc

emily-fields:


.nvmrc
```diff
@@ -0,0 +1 @@
+v18.20.6
```
wszeborowskimateusz, line 1: How about example apps? Would it work for them?

emily-fields: This works for me in example, example-vanilla and the no build example

wszeborowskimateusz: But did you open a new console or reuse the single one? I think to fully test it you would need to open the new console directly in the example app and then run npm i && npm run dev

emily-fields: Yeah I opened a new terminal and it worked for me. Is that not working for you?

wszeborowskimateusz (approved):

PR #104, opened 2025-03-21
Update table column widths

emily-fields:
image

### DIFF
--- a/src/sheet/components/SheetDataEditorCell.tsx
+++ b/src/sheet/components/SheetDataEditorCell.tsx
@@ -68,6 +68,7 @@ export default function SheetDataEditorCell({
         <div
           onClick={(e) => !readOnly && e.detail > 1 && setEditMode(true)}
           className={`h-full w-full py-4 pr-3 pl-4 ${cellBackgroundColor} truncate overflow-hidden whitespace-nowrap`}
+          title={`${nonEmptyValue}`}
         >
           {nonEmptyValue}
         </div>
--- a/src/sheet/components/SheetDataEditorCell.tsx
+++ b/src/sheet/components/SheetDataEditorCell.tsx
@@ -67,7 +67,7 @@ export default function SheetDataEditorCell({
       >
         <div
           onClick={(e) => !readOnly && e.detail > 1 && setEditMode(true)}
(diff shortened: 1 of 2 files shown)


src/sheet/components/SheetDataEditorCell.tsx
```diff
@@ -67,7 +67,7 @@ export default function SheetDataEditorCell({
       >
         <div
           onClick={(e) => !readOnly && e.detail > 1 && setEditMode(true)}
-          className={`h-full w-full py-4 pr-3 pl-4 ${cellBackgroundColor}`}
+          className={`h-full w-full py-4 pr-3 pl-4 ${cellBackgroundColor} truncate overflow-hidden whitespace-nowrap`}
```
wszeborowskimateusz, line 70: So if value exceeds the width we would not be able to read it fully? Or would it scroll somhow?

wszeborowskimateusz:
And what is the goal of this PR? What are we trying to achieve here?

src/sheet/components/SheetDataEditorCell.tsx
czhu12: Yeah truncate seems fine cause if you double click you can edit it right?

czhu12: the problem with not doing this is a single long value creates havoc on the UI

wszeborowskimateusz: Maybe we would also wanna set title={value} so that user can see the whole value on hover?

emily-fields:
We are trying to make the columns narrower if the content is short and then truncate the text if the content is long

src/sheet/components/SheetDataEditorCell.tsx
emily-fields: Hmm to me it seems like that isn't really necessary since they can click into it? But if you and Chris think we should include that I can

wszeborowskimateusz: What if column is readonly? Then they cannot click on it

czhu12 (approved):

PR #114, opened 2025-03-30
Add gray sections & refactor

emily-fields:


example/src/index.css
```diff
@@ -41,4 +41,14 @@ footer {
 		overflow: hidden;
     clip-path: polygon(0 0, 100% 0%, 115% 85%, 0 100%);
   }
+}
+
+.section {
```
wszeborowskimateusz: Why don't we use Tailwind for this?

wszeborowskimateusz: I think we still don't use Tailwind? I don';t think we need any custom css classes? Thats not the Tailwind way to do things?

example/src/App.tsx
```diff
@@ -46,7 +48,9 @@ export default function App() {
             </p>
           </div>
         </div>
+      </div>
 
+      <div className="section">
```
wszeborowskimateusz: I think those should be custom react components and not classes - I think that woukd be the Tailwind way to do things

example/src/App.tsx
```diff
@@ -14,20 +14,22 @@ export default function App() {
...
-          <h3 className="font-title">Installation</h3>
-          <p>
-            Works with any Javascript application,{' '}
-            <span className="font-semibold italic">
-              {"even if you don't use React."}
-            </span>
-          </p>
+      <div className="section">
+        <div className="content">
+          <div className="documentation-container">
```
wszeborowskimateusz: This should also probably just be inline tailwind css classes?

wszeborowskimateusz: Or some react component if we need to reuse it

wszeborowskimateusz (approved):

PR #127, opened 2025-04-07
Success page

emily-fields:


src/status/utils.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export function getTotalRows(sheetData: SheetState[]) {
+  return sheetData.reduce((total, sheet) => total + sheet.rows.length, 0);
+}
+
+function getCsv(rows: SheetRow[]) {
+  const headers = Object.keys(rows[0]);
+  const CsvContent = [
+    headers.join(','),
+    ...rows.map((row) => headers.map((h) => `"${row[h]}"`).join(',')),
```
czhu12: Can you test if this works if there is a cell with a " in it? I think you'll end up with improperly escaped quotes

emily-fields: I just tested this with ' and " and it worked for me

czhu12 (approved):

docs/v0.1.0/api-reference/importer-props.mdx
```diff
@@ -100,3 +106,14 @@ Sets the maximum allowed file size in bytes. Files exceeding this limit will be
 
 **Description:**  
 Provides a custom function to generate suggested mappings based on CSV headers.
+
+---
+
+## onSuccessRedirectUrl
```
wszeborowskimateusz, line 112: Shouldn't this be a callback? Wouldn't that be more customizable? Then integrator could just navigate to new page if they want to?

src/importer/types.ts
```diff
@@ -58,6 +60,7 @@ export interface ImporterState {
   rowFile?: File;
   columnMappings?: ColumnMapping[];
   importProgress: number;
+  importStatistics?: ImportStatistics;
```
wszeborowskimateusz, line 63: @emily-fields  I think you would also need to add it to the docs - under importer-state.mdx

PR #136, opened 2025-04-22
Add XSLX support

celina-lopez:
Screenshot 2025-04-22 at 12 27 12 PM


wszeborowskimateusz:
Seems like this one library doubles the size of the whole package?

taehee-kim-sherpas:
Can't we just add xlsx support as a kind of plugin or convert option props?

czhu12:
Addressed in https://github.com/HelloCSV/HelloCSV/pull/194

PR #142, opened 2025-04-23
add indexDB functionality

celina-lopez:
Save state to the browser's IndexDB. That way users can refresh the page while their state persists.


src/utils/storage.ts
```diff
@@ -0,0 +1,83 @@
...
+      }
+    };
+  });
+}
+
+export async function setIndexedDBState(state: ImporterState): Promise<void> {
+  return new Promise((resolve, reject) => {
+    const key = getStateKey(state.sheetDefinitions);
+    const value = { ...state } as any;
+    delete value.sheetDefinitions;
```
celina-lopez: sheetDefinitions have functions within, this should not be saved in the database

src/importer/index.tsx
```diff
@@ -37,11 +38,38 @@ function ImporterBody({
...
+    };
+    fetchState();
+  }, [sheets]);
+
+  if (initialState == null) {
+    return <div>Loading...</div>;
+  }
 
-  const [state, dispatch] = useReducer(reducer, buildInitialState(sheets));
+  // eslint-disable-next-line react-hooks/rules-of-hooks
```
wszeborowskimateusz: Why do we need this here and in the line below?

src/status/components/SummaryInfo.tsx
```diff
@@ -16,7 +16,7 @@ type Props = {
   sheetData: SheetState[];
   mode: Mode;
   statistics?: ImportStatistics;
-  rowFile?: File;
+  rowFile?: { name: string; size: number; content?: any };
```
wszeborowskimateusz: Why did we make such a change? Isn't `File` built in type?

src/importer/reducer.ts
```diff
@@ -48,129 +68,117 @@ const reducer = (
...
+    };
+  } else if (
+    ['FAILED', 'PREVIEW', 'MAPPING', 'SUBMIT', 'UPLOAD'].includes(action.type)
+  ) {
+    newState = { ...state, mode: action.type.toLowerCase() as ImporterMode };
+  } else if (action.type === 'RESET') {
+    newState = buildInitialState(state.sheetDefinitions);
   }
+
+  setIndexedDBState(newState);
```
wszeborowskimateusz: Why do we need this bit? Isn't it already handled by useEffect above?

src/importer/reducer.ts
```diff
@@ -30,6 +32,24 @@ function recalculateCalculatedColumns(
...
+    const state = await getIndexedDBState(sheetDefinitions);
+    if (state != null) {
+      return state;
+    }
+
+    const newState = buildInitialState(sheetDefinitions);
+    setIndexedDBState(newState);
+    return newState;
+  } catch (error) {
+    console.error(error);
```
wszeborowskimateusz: I don't think we want to have console logs?

src/importer/reducer.ts
wszeborowskimateusz: Actually, maybe it would be better to write a custom usePersistedReducer hook, that would have built it useEffect to store date in the db?

src/importer/index.tsx
wszeborowskimateusz: Generally it looks like a code smell, I don't think hooks can be called conditionally in react - like you do it here

src/importer/index.tsx
```diff
@@ -91,6 +108,25 @@ function ImporterBody({
...
+          reader.readAsDataURL(file);
+          reader.onload = function () {
+            resolve(reader.result);
+          };
+        });
+
+        dispatch({
+          type: 'FILE_UPLOADED',
+          payload: {
+            rowFile: {
```
wszeborowskimateusz: Well, it doesn't look like row file anymore?

src/importer/index.tsx
```diff
@@ -37,11 +38,38 @@ function ImporterBody({
...
+    if (isInitialRender.current) {
+      isInitialRender.current = false;
+      return;
+    }
+
+    targetRef.current?.scrollIntoView({ behavior: 'smooth' });
+  }, [state?.mode]);
+
+  if (!state || !dispatch) {
+    return <div>Loading...</div>;
```
wszeborowskimateusz: When can state or dispatch be nil?

src/importer/reducer.ts
```diff
@@ -48,129 +68,117 @@ const reducer = (
...
+    newState = { ...state, mode: action.type.toLowerCase() as ImporterMode };
+  } else if (action.type === 'RESET') {
+    newState = buildInitialState(state.sheetDefinitions);
   }
+
+  setIndexedDBState(newState);
+  return newState;
 };
 
 export { reducer, buildInitialState };
```
wszeborowskimateusz: Why do we export buildInitialState? Do we use it outside of this file?

src/utils/storage.ts
wszeborowskimateusz, line 1: Maybe this file should be located in importer/utils.ts? This seems more consistent with other features?

src/utils/storage.ts
```diff
@@ -0,0 +1,83 @@
...
+      if (!db.objectStoreNames.contains(STORE_NAME)) {
+        db.createObjectStore(STORE_NAME);
+      }
+    };
+  });
+}
+
+function getStateKey(array: any[]): string {
+  const key = array.map((item) => `${item.id}-${item.label}`).join('|');
+  return `importer-state-${hashString(key)}`;
```
wszeborowskimateusz: Why do we need to hash it?

src/utils/storage.ts
wszeborowskimateusz: Perhaps this could be a comment in the code?

src/utils/storage.ts
```diff
@@ -0,0 +1,83 @@
+import { ImporterState } from '../importer/types';
+import { SheetDefinition } from '../sheet/types';
+
+const DB_NAME = 'HelloCSV';
```
wszeborowskimateusz: How does it work when app integrates the plugin twice - like renders 2 different importers? Does each importer have its own db?

src/utils/storage.ts
```diff
@@ -0,0 +1,83 @@
...
+          resolve(result);
+        } catch (error) {
+          reject(error);
+        }
+      };
+    };
+
+    request.onupgradeneeded = (event) => {
+      const db = (event.target as IDBOpenDBRequest).result;
+      if (!db.objectStoreNames.contains(STORE_NAME)) {
```
wszeborowskimateusz: How about scenarios when we bump version because of some schema changes? Shouldn't we invalidate (delete) data from the DB with the old version? I think currently nothing would happen and the app might load state with old schema and cause the app to crash

src/utils/storage.ts
```diff
@@ -0,0 +1,83 @@
...
+      const db = request.result;
+      const transaction = db.transaction(STORE_NAME, 'readonly');
+      const store = transaction.objectStore(STORE_NAME);
+      const getRequest = store.get(key);
+
+      getRequest.onerror = () => reject(getRequest.error);
+      getRequest.onsuccess = () => {
+        try {
+          const result = getRequest.result;
+          result.sheetDefinitions = sheetDefinitions;
```
wszeborowskimateusz: I am wondering if this won't cause some problems - like what if e.g. integrator changes some definition (e.g. id of some sheet)? We store currentSheetId in the state and if integrator changes the id in the definition, then won't that cause the app to crash, since sheet with currentSheetId doesn't exist?

src/importer/reducer.ts
celina-lopez: we do, we use it in the index.tsx file

src/utils/storage.ts
celina-lopez: no this should have the same database, just different key,values in said database

celina-lopez: in the indexDBConfig, they could add a custom key prefix if needed. (if they happen to have two importers with the exact sheetDefinitions)

The key in the indexDB is a hash generated from the sheetDefinition

src/utils/storage.ts
celina-lopez: if the integrator changes a sheetDefinition, we would create a new row in the indexDB database

src/importer/index.tsx
celina-lopez: removed

src/status/components/SummaryInfo.tsx
celina-lopez: We can't save Files in the indexDB
We can however save the content of the file, along with the name and size.

src/importer/reducer.ts
celina-lopez: This is to save to the indexDB when the state is updated

src/importer/index.tsx
celina-lopez: removed

src/importer/reducer.ts
celina-lopez: we do when we first build the state in `importer/index.tsx`

src/utils/storage.ts
celina-lopez: To make a custom key in our indexDB

if they update the sheetsDefinition, then we won't load the old data

src/utils/storage.ts
celina-lopez: updated the code to remove old data if the version changes

src/status/components/SummaryInfo.tsx
wszeborowskimateusz: Are you sure about that? I just tested it and it seems like indexDB can store the actual file

Screenshot 2025-05-07 at 7 19 34 AM

src/importer/reducer.ts
wszeborowskimateusz: Yes, but I think it would be cleaner if this is handled via useEffect in a custom hook - this way
1. We can just set a new state
2. Works automatically for all actions
3. We can hide setting initial state from consumer

src/importer/index.tsx
wszeborowskimateusz: As above - it does seem like we can store the rowFile in the DB

src/importer/reducer.ts
wszeborowskimateusz: I think this could be hidden with custom usePersistedReducer hook proposed above

src/utils/storage.ts
wszeborowskimateusz: Yes, but you hash a string consisting of id and labels. Whats the difference between this string and a hash. Isn't it the case that when they update sheet definition the string would also update?

src/utils/storage.ts
wszeborowskimateusz: Is the `if (oldVersion < DB_VERSION) {` needed there? From looking at the docs this event is fired when `an attempt was made to open a database with a version number higher than its current version.`

example/src/App.tsx
```diff
@@ -94,6 +95,10 @@ export default function App() {
         <EmployeeSheetImporter />
       </Section>
 
+      <Section className="bg-gray-100">
```
wszeborowskimateusz: I am wondering if this just doesn't complicate the demo page? Shouldn't we simply add indexDB support to other 2 examples we have?

src/importer/index.tsx
```diff
@@ -53,6 +64,8 @@ function ImporterBody({
     importProgress,
     importStatistics,
   } = state;
+  console.log('mode state', state.mode);
+  console.log('mode', mode);
```
wszeborowskimateusz: I think we should remove those console logs

src/importer/index.tsx
```diff
@@ -35,13 +35,24 @@ function ImporterBody({
...
 
   const isInitialRender = useRef(true);
   const targetRef = useRef<HTMLDivElement | null>(null);
+  const [state, dispatch] = useReducer(
+    reducer,
+    buildInitialState(sheets, indexDBConfig)
+  );
 
-  const [state, dispatch] = useReducer(reducer, buildInitialState(sheets));
+  useEffect(() => {
```
wszeborowskimateusz: I think this should be a part of usePersistedReducer custom hook

src/importer/reducer.ts
```diff
@@ -30,7 +33,10 @@ function recalculateCalculatedColumns(
   return row;
 }
 
-function buildInitialState(sheetDefinitions: SheetDefinition[]): ImporterState {
+function buildInitialState(
+  sheetDefinitions: SheetDefinition[],
+  indexDBConfig: IndexDBConfig = { enabled: false }
```
wszeborowskimateusz: 1. I think the default should be defined in ImporterBody
2. Shouldn't it be turned on by default?

src/importer/storage.ts
```diff
@@ -0,0 +1,96 @@
...
+        }
+      };
+    };
+
+    request.onupgradeneeded = (event) => {
+      const db = (event.target as IDBOpenDBRequest).result;
+      const oldVersion = event.oldVersion;
+
+      // Clear data if upgrading from an older version
+      if (oldVersion < DB_VERSION) {
```
wszeborowskimateusz: Is this if needed?

src/importer/storage.ts
```diff
@@ -0,0 +1,96 @@
...
+    request.onupgradeneeded = (event) => {
+      const db = (event.target as IDBOpenDBRequest).result;
+      if (!db.objectStoreNames.contains(STORE_NAME)) {
+        db.createObjectStore(STORE_NAME);
+      }
+    };
+  });
+}
+
+function stateKey(array: any[], customKey?: string | null): string {
```
wszeborowskimateusz: I think array should have a type of `SheetDefinition[]` and probably be called sheetDefinitions

src/importer/storage.ts
```diff
@@ -0,0 +1,96 @@
...
+  const prefix = customKey ? `importer-state-${customKey}` : 'importer-state';
+  const key = array.map((item) => `${item.id}-${item.label}`).join('|');
```
wszeborowskimateusz: Shouldn't we do something like `JSON.stringify(sheetDefinitions)`?

src/importer/types.ts
```diff
@@ -57,10 +63,11 @@ export interface ImporterState {
   validationErrors: ImporterValidationError[];
   sheetData: SheetState[];
   parsedFile?: ParsedFile;
-  rowFile?: File;
+  fileData?: FileData;
   columnMappings?: ColumnMapping[];
   importProgress: number;
   importStatistics?: ImportStatistics;
+  indexDBConfig?: IndexDBConfig;
```
wszeborowskimateusz: Why do we need to store it in the state?

wszeborowskimateusz:
I am proposing something like this for the reducer

```ts
import { applyTransformations } from '../transformers';
import {
  CellChangedPayload,
  ImporterAction,
  ImporterMode,
  ImporterState,
  IndexDBConfig,
  SheetDefinition,
  SheetRow,
} from '../types';
import { getIndexedDBState, setIndexedDBState } from './storage';
import { applyValidations } from '../validators';
import { Dispatch, useEffect, useReducer } from 'preact/hooks';

function recalculateCalculatedColumns(
  row: SheetRow,
  payload: CellChangedPayload,
  state: ImporterState
): SheetRow {
  const sheetDefinition = state.sheetDefinitions.find(
    (s) => s.id === payload.sheetId
  );

  if (sheetDefinition != null) {
    const calculatedColumns = sheetDefinition.columns.filter(
      (column) => column.type === 'calculated'
    );

    calculatedColumns.forEach((column) => {
      row[column.id] = column.typeArguments.getValue(row);
    });
  }

  return row;
}

function buildInitialState(sheetDefinitions: SheetDefinition[]): ImporterState {
  return {
    sheetDefinitions,
    currentSheetId: sheetDefinitions[0].id,
    mode: 'upload',
    validationErrors: [],
    sheetData: sheetDefinitions.map((sheet) => ({
      sheetId: sheet.id,
      rows: [],
    })),
    importProgress: 0,
  };
}

async function buildStateWithIndexedDB(
  sheetDefinitions: SheetDefinition[],
  indexDBConfig: IndexDBConfig
): Promise<ImporterState> {
  try {
    const state = await getIndexedDBState(
      sheetDefinitions,
      indexDBConfig.customKey
    );

    if (state != null) {
      return state;
    }

    const newState = buildInitialState(sheetDefinitions);
    setIndexedDBState(newState, indexDBConfig.customKey);
    return newState;
  } catch (_error) {
    return buildInitialState(sheetDefinitions);
  }
}

const reducer = (
  state: ImporterState,
  action: ImporterAction
): ImporterState => {
  if (action.type === 'ENTER_DATA_MANUALLY') {
    const emptyData = state.sheetDefinitions.map((sheet) => ({
      sheetId: sheet.id,
      rows: Array.from(
        { length: action.payload.amountOfEmptyRowsToAdd },
        () => ({})
      ),
    }));
    return { ...state, mode: 'preview', sheetData: emptyData };
  } else if (action.type === 'FILE_PARSED') {
    return {
      ...state,
      parsedFile: action.payload.parsed,
      fileData: action.payload.fileData,
      mode: 'mapping',
    };
  } else if (action.type === 'COLUMN_MAPPING_CHANGED') {
    return {
      ...state,
      columnMappings: action.payload.mappings,
    };
  } else if (action.type === 'DATA_MAPPED') {
    return {
      ...state,
      sheetData: applyTransformations(
        state.sheetDefinitions,
        action.payload.mappedData
      ),
      mode: 'preview',
      validationErrors: applyValidations(
        state.sheetDefinitions,
        action.payload.mappedData
      ),
    };
  } else if (action.type === 'CELL_CHANGED') {
    const newData = state.sheetData.map((sheet) => {
      if (sheet.sheetId === action.payload.sheetId) {
        const newRows = [...sheet.rows];

        newRows[action.payload.rowIndex] = recalculateCalculatedColumns(
          action.payload.value,
          action.payload,
          state
        );

        return { ...sheet, rows: newRows };
      } else {
        return sheet;
      }
    });

    return {
      ...state,
      sheetData: applyTransformations(state.sheetDefinitions, newData),
      validationErrors: applyValidations(state.sheetDefinitions, newData),
    };
  } else if (action.type === 'REMOVE_ROWS') {
    const newData = state.sheetData.map((sheet) => {
      if (sheet.sheetId === action.payload.sheetId) {
        return {
          ...sheet,
          rows: sheet.rows.filter((row) => !action.payload.rows.includes(row)),
        };
      }

      return sheet;
    });

    return {
      ...state,
      sheetData: newData,
      validationErrors: applyValidations(state.sheetDefinitions, newData),
    };
  } else if (action.type === 'ADD_EMPTY_ROW') {
    const newData = state.sheetData.map((data) => {
      if (data.sheetId !== state.currentSheetId) {
        return data;
      }

      return {
        ...data,
        rows: [...data.rows, {}],
      };
    });

    return {
      ...state,
      sheetData: newData,
    };
  } else if (action.type === 'SHEET_CHANGED') {
    return {
      ...state,
      currentSheetId: action.payload.sheetId,
    };
  } else if (action.type === 'PROGRESS') {
    return {
      ...state,
      importProgress: action.payload.progress,
    };
  } else if (action.type === 'COMPLETED') {
    return {
      ...state,
      mode: 'completed',
      importStatistics: action.payload.importStatistics,
    };
  } else if (
    ['FAILED', 'PREVIEW', 'MAPPING', 'SUBMIT', 'UPLOAD'].includes(action.type)
  ) {
    return {
      ...state,
      mode: action.type.toLowerCase() as ImporterMode,
    };
  } else if (action.type === 'RESET') {
    return buildInitialState(state.sheetDefinitions);
  } else if (action.type === 'SET_STATE') {
    return action.payload.state;
  }

  return state;
};

export function usePersistedReducer(
  sheetDefinitions: SheetDefinition[],
  indexDBConfig: IndexDBConfig
): [ImporterState, Dispatch<ImporterAction>] {
  const [state, dispatch] = useReducer(
    reducer,
    buildInitialState(sheetDefinitions)
  );

  useEffect(() => {
    const fetchState = async () => {
      if (indexDBConfig.enabled) {
        const newState = await buildStateWithIndexedDB(
          sheetDefinitions,
          indexDBConfig
        );
        dispatch({ type: 'SET_STATE', payload: { state: newState } });
      }
    };
    fetchState();
  }, [sheetDefinitions, indexDBConfig]);

  useEffect(() => {
    if (indexDBConfig.enabled) {
      setIndexedDBState(state, indexDBConfig.customKey);
    }
  }, [state, indexDBConfig]);

  return [state, dispatch];
}

```

src/importer/reducer.ts
celina-lopez: 1. I'll update
2. I'm not sure if we should automatically save to their indexDB. A single device such as a mac could be used by multiple people. Maybe data that is upload using one person's credentials would be seen by someone else using the same device. This feature is probably better off something that is manually turned on instead.

wszeborowskimateusz: 2. I think that makes sense.  How would that work for labconnect though? Would integrator need to add a customKey and store userId in it so that different users would have different stores?

wszeborowskimateusz:
When you add a new key to public interface / state you also need to update the docs - `/docs` folder

example/src/App.tsx
czhu12: Yes please!

src/importer/storage.ts
celina-lopez: just in case we upgrade the DB for any reason 🤷🏼

wszeborowskimateusz: Yes, but what would be the case that this method is called and `oldVersion >= DB_VERSION`? Is it possible?

celina-lopez: hmm maybe you're right

I'll remove this and only add something else if we do upgrade our database

wszeborowskimateusz: Just for clarity - I think we can remove the `if` - but leave it's body - just call it anytime

wszeborowskimateusz: I think you removed too much 😓 - we still want to recreate the DB on each version upgrade IMO

src/status/components/SummaryInfo.tsx
celina-lopez: oh what.... let me check again
It was erroring for me when I first started

src/importer/reducer.ts
celina-lopez: @czhu12 Idk what your vision is for this, but would we just leave it up to the user to figure out the logic?

src/importer/types.ts
celina-lopez: so we can see in the reducer if we need to save it in the indexDB

wszeborowskimateusz: But we could pass it as props - see my proposal here - https://github.com/HelloCSV/HelloCSV/pull/142#issuecomment-2858932498

celina-lopez:
> When you add a new key to public interface / state you also need to update the docs - `/docs` folder

@czhu12 mentioned this was automatic, does a command need to be run for that to happen? Or is this manual work?

wszeborowskimateusz:
> > When you add a new key to public interface / state you also need to update the docs - `/docs` folder
> 
> @czhu12 mentioned this was automatic, does a command need to be run for that to happen? Or is this manual work?

Whats automatic is deployment of the docs - not writing them

example/src/App.tsx
celina-lopez: removed

src/importer/storage.ts
celina-lopez: shouldn't we add this functionality when we do need to upgrade? Right now it will always be on version 1 unless a new PR upgrades the DB version

wszeborowskimateusz: I think we could add it already. 
1. Don't you need to have it for the initial version when DB doesn't exist yet on the user end?
2. I think the implementation would always be the same  - bumping version means that there is some breaking change in the state (I don't think we will bump version for minor state changes - we could write in in `deployment.mdx` file - to make sure DB_VERSION is bumped in case of breaking changes in ImporterState) and hence the db should be dropped - we don't want to risk having incompatible types in there

src/status/components/SummaryInfo.tsx
wszeborowskimateusz: Well, you see on the SS that it is possible. StackOverflow also says it should be possible. Perhaps there was some different error?

src/importer/types.ts
wszeborowskimateusz: Plus my proposal also saves state to DB as useEffect instead of doing it in the reducer - I think thats a cleaner solution. Reducer should only generate a new state - it shouldn't have any side effects - especially not async ones

src/importer/reducer.ts
wszeborowskimateusz: If we leave the problem to integrator we should definitely describe that in the docs

src/importer/types.ts
celina-lopez: added updates to the file

celina-lopez:
@wszeborowskimateusz gotcha will do

src/status/components/SummaryInfo.tsx
celina-lopez: ok, i was able to add.
I'm not sure why it wasn't saving before, must have been some other bug.

src/importer/reducer.ts
```diff
@@ -168,9 +206,32 @@ const reducer = (
...
+  useEffect(() => {
+    const fetchState = async () => {
+      const newState = await buildState(sheets, indexDBConfig);
+      dispatch({ type: 'SET_STATE', payload: { state: newState } });
+    };
+    fetchState();
+  }, [sheets, indexDBConfig]);
+
+  useEffect(() => {
+    setIndexedDBState(state, indexDBConfig.customKey);
```
wszeborowskimateusz: Shouldn't we call that only when indexDB is enabled?

src/utils/storage.ts
celina-lopez: I mean we can just stringify the sheetDefinitions and leave it at that... but that would make for an awkward id

wszeborowskimateusz: ok, you convinced me - let's have it hashed

celina-lopez:
docs pr: https://github.com/HelloCSV/HelloCSV/pull/151

src/importer/storage.ts
```diff
@@ -0,0 +1,91 @@
...
+  const key = JSON.stringify(sheetDefinitions);
+  return `${prefix}-${hashString(key)}`;
+}
+
+function hashString(str: string): number {
+  let hash = 0;
+  for (let i = 0; i < str.length; i++) {
+    const char = str.charCodeAt(i);
+    hash = (hash << 5) - hash + char;
+    hash = hash & hash;
```
wszeborowskimateusz, line 88: What does this line do? Isn't it no-op?

celina-lopez: this is to make sure it stays within the 32-bit integer range

wszeborowskimateusz: ChatGPT says it doesn't work

Screenshot 2025-05-07 at 1 31 36 PM

wszeborowskimateusz (approved):

PR #146, opened 2025-04-30
Mobile view improvements

wszeborowskimateusz:
Changes
- Hide data preview for the mapper on mobile - I don't think there is place for it and it's not that crucial
- Fix buttons on mobile device for confirmation modal
- Fix editing cells on mobile device - now they would need to long press the cell to edit it

Otherwise, I think app looks ok on mobile

### DIFF
--- a/src/components/Button.tsx
+++ b/src/components/Button.tsx
@@ -13,7 +13,7 @@ interface Props {
   variant?: ButtonVariant;
   outline?: boolean;
   disabled?: boolean;
-  size?: 'xs' | 'sm' | 'md' | 'lg' | 'xl';
+  size?: 'xs' | 'sm' | 'md' | 'lg' | 'xl' | 'full';
   onClick?: () => void;
   style?: CSSProperties;
 }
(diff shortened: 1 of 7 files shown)


src/components/Button.tsx
emily-fields, line 1: One thing I was thinking is maybe we should get rid of the button variants that we aren't using. I added more than we needed initially because I thought we would use them but we aren't. We could do that in a future PR though?

wszeborowskimateusz: You talk about size variants?

wszeborowskimateusz: Or also button variant - like primary, secondary, etc?

emily-fields: Yeah. It isn't really related to this PR so we could just note it and do it in the future 😆

wszeborowskimateusz: I mean, I could remove them here

emily-fields: Your call 😈

wszeborowskimateusz: I did remove the size variant

emily-fields (approved):

PR #172, opened 2025-05-12
Allow using package in React apps

wszeborowskimateusz:


example/src/App.tsx
emily-fields, line 1: Maybe we should call this folder example-preact instead of example since now we have example-react?

wszeborowskimateusz: done

czhu12 (approved):

PR #179, opened 2025-05-13
Sync production to main

emily-fields:


.github/workflows/sync-production-to-main.yml
```diff
@@ -0,0 +1,33 @@
...
+    steps:
+      - name: Checkout main branch
+        uses: actions/checkout@v4
+        with:
+          ref: main
+
+      - name: Reset branch
+        run: |
+          git fetch origin production:production
+          git reset --hard production
```
wszeborowskimateusz, line 20: wait, so we create a branch from main or production?

wszeborowskimateusz:
Shouldn't we add a step to deployment guide?

docs/common/development/deployment.mdx
```diff
@@ -33,4 +33,6 @@ For major version release
 
 8. Create a new release in Github - https://github.com/HelloCSV/HelloCSV/releases. Remember about creating a tag (`v{x.y.z}`) and writing release notes.
 
-9. Merge `main` to `production` branch - this would trigger docs deployment to Mintlify.
```
wszeborowskimateusz: I don't thinbk this file exists on main?

.github/workflows/sync-production-to-main.yml
```diff
@@ -0,0 +1,33 @@
+name: Sync production to main
+
+on:
```
wszeborowskimateusz, line 3: do we want to also add workflow here to be able to dispatch it manually?

.github/workflows/sync-production-to-main.yml
emily-fields: https://github.com/peter-evans/create-pull-request/blob/main/docs/examples.md

image

.github/workflows/sync-production-to-main.yml
emily-fields: Why do you think we should include that?

.github/workflows/sync-production-to-main.yml
wszeborowskimateusz: Isn't it the other way round in the example?

.github/workflows/sync-production-to-main.yml
wszeborowskimateusz: I guess it doesn't hurt much, but maybe it could be useful sometimes 🤷

docs/v0.1.4/development/deployment.mdx
```diff
@@ -33,4 +33,6 @@ For major version release
 
 8. Create a new release in Github - https://github.com/HelloCSV/HelloCSV/releases. Remember about creating a tag (`v{x.y.z}`) and writing release notes.
 
-9. Merge `main` to `production` branch - this would trigger docs deployment to Mintlify.
+9. Merge `main` to `production` branch - this would trigger docs deployment to Mintlify
+
+10. Merge the automated PR 'Sync production to main' to keep the `main` branch in sync with the `production` branch
```
wszeborowskimateusz, line 38: I think if we put it there we should also put it in v.0.2.0 docs

.github/workflows/sync-production-to-main.yml
emily-fields: Yes, because the example is doing the opposite of what we are doing

.github/workflows/sync-production-to-main.yml
emily-fields: I don't think it is necessary because this runs anytime there is a push to production?

wszeborowskimateusz: yeah, but the action could fail. But I guess we could rerun it then 🤷‍♂️

.github/workflows/sync-production-to-main.yml
wszeborowskimateusz: so we create branch from production or main?

wszeborowskimateusz (approved):

PR #183, opened 2025-05-15
Release v0.3.0

celina-lopez:
- change indexDBConfig -> persistenceConfig


docs/docs.json
```diff
@@ -15,34 +15,34 @@
...
-              "v0.2.0/api-reference/importer-props",
-              "v0.2.0/api-reference/sheet-definition-props",
-              "v0.2.0/api-reference/column-types",
-              "v0.2.0/api-reference/common-column-props",
-              "v0.2.0/api-reference/importer-state",
-              "v0.2.0/api-reference/sheet-state",
-              "v0.2.0/api-reference/validators",
-              "v0.2.0/api-reference/transformers",
-              "v0.2.0/api-reference/importer-validation-error"
+              "v0.3.0/api-reference/importer-props",
```
wszeborowskimateusz, line 32: Since thats a major release, you should create a new version entry and write a migration guide

docs/v0.1.4/api-reference/importer-props.mdx
```diff
@@ -121,7 +121,7 @@ Callback executed when the user has finished reviewing the import summary. This
 
 ---
 
-## indexDBConfig
+## persistenceConfig
```
wszeborowskimateusz: This shouldn't be changed in 0.1.4, since the field is called indexDBConfig in 0.1.4

docs/v0.1.4/get-started/usage.mdx
```diff
@@ -95,7 +95,7 @@ import 'hello-csv/peer/index.css';
   onComplete={(data) => {
     console.log(data);
   }}
-  indexDBConfig={{ enabled: true }}
+  persistenceConfig={{ enabled: true }}
```
wszeborowskimateusz: as above

example/dist/android-chrome-192x192.png
wszeborowskimateusz, line 1: The dist folder should be git ignored - I think it already is? But you shouldn't commit those files

.gitignore
```diff
@@ -22,3 +22,5 @@ dist-ssr
 *.njsproj
 *.sln
 *.sw?
+
+example/dist/*
```
wszeborowskimateusz, line 26: I don't understand why this is needed 🤔

celina-lopez: didn't you say not to commit the example/dist/* files?

wszeborowskimateusz: Yeah, but I think it is already gitignored in example's gitignore

wszeborowskimateusz: Plus I don't think we have example folder? I guess we have example-preact, example-react and example-vanilla?

docs/docs.json
wszeborowskimateusz: I saw you created a new files for v0.3.0. Thats good, but you also need to add a new entry in this file - not just replace 0.2.0 with 0.3.0. Look at how we have here 0.1.4 and 0.2.0. So we should maintain those and just add a new entry

docs/v0.3.0/upgrade/from-v-0-1-4.mdx
wszeborowskimateusz, line 1: Looks like the name of the file is not correct? I guess it should be from-v-0-2-0?

czhu12 (approved):

PR #192, opened 2025-05-19
Custom translation resource support

taehee-kim-sherpas:
## Problem

We want to use HelloCSV in our product, but it currently doesn't support the `ko` (Korean 한국어) translation.

But... the more languages this library supports, the larger its bundle size becomes.
And... users of the library might want to customize labels and messages to match their brand tone.

## Solution

We would like to provide custom translation resources via props.

https://github.com/HelloCSV/HelloCSV/blob/main/src/i18/index.tsx

But what happens if a translation key is added in a future version and we miss it?

We can use a stronger type definition (the `Translation` type) to catch missing keys (e.g., `importFailed` and `failedDescription` in the `fr` translation).

### DIFF
--- a/src/i18/index.tsx
+++ b/src/i18/index.tsx
@@ -83,7 +83,7 @@ export function TranslationProvider({
 }: {
   children: ReactNode;
   selectedLocale?: string;
-  translationResources?: Resources;
+  translationResources?: Record<string, Translation>;
 }) {
   const locale = selectedLocale ?? defaultLocale;
 
--- a/src/i18/index.tsx
+++ b/src/i18/index.tsx
@@ -11,20 +11,32 @@ import {
   TranslationContextType,
 } from './types';
 
-/// Writing a custom translations module to reduce the size of the package
+type Prettify<T> = {
+  [K in keyof T]: T[K] extends object ? Prettify<T[K]> : T[K];
+} & unknown;
 
-const resources: Resources = {
+type CreateTranslation<T> = {
+  [K in keyof T]: T[K] extends string
+    ? string
+    : T[K] extends object
+      ? CreateTranslation<T[K]>
+      : T[K];
+};
+
+export type Translation = Prettify<CreateTranslation<typeof enTranslation>>;
+
+const resources: Record<string, Translation | undefined> = {
   en: enTranslation,
   fr: frTranslation,
   'pt-BR': ptBRTranslation,
-};
+} satisfies Resources;
 
 const defaultLocale = 'en';
 
-function extractTranslation(currentLocale: string, key: string) {
+function extractTranslation(currentLocale: string, key: string, translationResources?: Resources) {
   const keyParts = key.split('.');
 
-  let result: any = resources[currentLocale];
+  let result: any = translationResources?.[currentLocale] ?? resources[currentLocale];
 
   for (const k of keyParts) {
     if (result && typeof result === 'object') {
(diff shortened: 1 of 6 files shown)


src/importer/types.ts
```diff
@@ -84,29 +86,29 @@ export interface RemoveRowsPayload {
...
+    payload: { parsed: ParsedFile; rowFile: File };
+  } // Sets the parsed file and changes the mode to 'mapping'
   | { type: 'UPLOAD' } // Changes the mode to 'upload' - used when going back from in the mapping screen
   | { type: 'COLUMN_MAPPING_CHANGED'; payload: { mappings: ColumnMapping[] } } // Sets the proper mappings
   | { type: 'DATA_MAPPED'; payload: { mappedData: MappedData } } // Sets mapped data as sheetData, optionally runs onDataColumnsMapped callback calls validations, changes the mode to 'preview'
   | {
-      type: 'CELL_CHANGED';
-      payload: CellChangedPayload;
-    } // Searches for the cell and changes the value, calls validations
+    type: 'CELL_CHANGED';
```
wszeborowskimateusz: Seems like you do not use prettier? Could you reformat the file with prettier?

src/importer/types.ts
```diff
@@ -1,3 +1,4 @@
+import { Translation } from '../i18';
```
wszeborowskimateusz: I think for consistency this import should come from ../types as all other type imports in this file

src/i18/index.tsx
```diff
@@ -11,20 +11,32 @@ import {
...
+
+type CreateTranslation<T> = {
+  [K in keyof T]: T[K] extends string
+    ? string
+    : T[K] extends object
+      ? CreateTranslation<T[K]>
+      : T[K];
+};
+
+export type Translation = Prettify<CreateTranslation<typeof enTranslation>>;
```
wszeborowskimateusz: Perhaps for consistency those could be added to types file? (i18/types.ts)

src/importer/types.ts
```diff
@@ -15,6 +16,7 @@ import {
 
 export interface ImporterDefinition {
   sheets: SheetDefinition[];
+  translationResources?: Record<string, Translation>;
```
wszeborowskimateusz, line 19: I think you should also update docs/v0.3.0/api-reference/importer-props and include the new prop

czhu12: I'm happy to update the docs after this gets merged!

src/importer/types.ts
taehee-kim-sherpas: Yes, I use biome. I reformatted it

src/importer/types.ts
taehee-kim-sherpas: Done!

src/i18/index.tsx
taehee-kim-sherpas: I moved them to types.ts

src/importer/types.ts
taehee-kim-sherpas: I updated the docs.

src/importer/types.ts
```diff
@@ -9,12 +9,14 @@ import {
...
   SheetRow,
   ImportStatistics,
+  Translation,
 } from '../types';
 
 // --------- Importer Definition Types ---------
 
 export interface ImporterDefinition {
   sheets: SheetDefinition[];
+  translationResources?: Record<string, Translation>;
```
wszeborowskimateusz, line 19: It does seem that type definition here (Record<string, Translation>) and in TranslationProvider (Resources) is different. Any reason for it? Shouldn't those be the same?

taehee-kim-sherpas: No problem. I changed it.

wszeborowskimateusz (approved):

wszeborowskimateusz:
Seems like format check failed?

Screenshot 2025-05-19 at 6 20 34 PM

taehee-kim-sherpas:
`npx prettier --write  "src/**/*.tsx" "src/**/*.ts"` -> done

PR #193, opened 2025-05-19
fix: use defaultValue for inline editing

taehee-kim-sherpas:
## Problem

We can't use inline editing feature in preview with `react`

## Solution

use `defaultValue` prop instead of `value`

https://react.dev/reference/react-dom/components/input#providing-an-initial-value-for-an-input

### DIFF
--- a/src/components/Input.tsx
+++ b/src/components/Input.tsx
@@ -50,7 +50,7 @@ const Input = forwardRef<HTMLInputElement, Props>(
           type={type}
           inputMode={type === 'number' ? 'numeric' : 'text'}
           placeholder={placeholder}
-          value={value}
+          defaultValue={value}
           onChange={(e) => onChange?.(getParsedValue(e))}
           className={`${classes} ${iconBuilder != null ? 'pl-10' : ''} ${clearable ? 'pr-10' : ''} focus:outline-hello-csv-primary col-start-1 row-start-1 block rounded-md bg-white px-3 py-1.5 text-base text-gray-900 outline-1 -outline-offset-1 outline-gray-300 placeholder:text-gray-400 focus:outline-2 focus:-outline-offset-2 sm:text-sm/6`}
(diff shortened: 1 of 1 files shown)


src/components/Input.tsx
```diff
@@ -50,7 +50,7 @@ const Input = forwardRef<HTMLInputElement, Props>(
           type={type}
           inputMode={type === 'number' ? 'numeric' : 'text'}
           placeholder={placeholder}
-          value={value}
+          defaultValue={value}
```
wszeborowskimateusz, line 53: @taehee-kim-sherpas could you specify the details? I just tested it with react (with value prop) and it works for me

wszeborowskimateusz: Tested from example/react app - that is with React 19.1.0

wszeborowskimateusz: ok, there seems to be something wrong, you are right

taehee-kim-sherpas:
Test case for react csv preview

Before
cell with no existing value -> can edit
cell with existing value -> can't edit

After
cell with no existing value -> can edit
cell with existing value -> can edit

wszeborowskimateusz:
> Test case for react csv preview
> 
> Before cell with no existing value -> can edit cell with existing value -> can't edit
> 
> After cell with no existing value -> can edit cell with existing value -> can edit

Yeah, I can see that, I am just thinking if there aren't any cases where using defaultValue won't work, because e.g. we want to set the value from the outside

wszeborowskimateusz:
I am wondering if it wouldn't be safer to do somthing like this

Screenshot 2025-05-19 at 6 58 30 PM

wszeborowskimateusz:
https://github.com/HelloCSV/HelloCSV/pull/197

wszeborowskimateusz:
Closing this one, as it should be fixed in #197

Issue #199, opened 2025-05-20
using virtual list


taehee-kim-sherpas:
## Problem

When i upload more than 2000 rows (with 30 colums) csv, It takes long time to render (more than 10s). (it's only 1mb)

## Solution
Use a virtual list—in other words, render only the rows currently visible to the user.

### [@tanstack/virtual](https://tanstack.com/virtual/latest/docs/introduction)
It doesn't work well with traditional <table> layouts—we would need to use grid or flex instead.
Column widths must be set manually.

4.6Minified + Gzipped

### [Virtuoso Table](https://virtuoso.dev/hello-table/)
Allows us to keep the existing table layout.

18.1 Minified + Gzipped

If I get permission, I’d like to work on this and submit a PR.

czhu12:
Really great idea, would love to see something like this! what kinds of permissions do you need?

czhu12:
I think the only challenge with Virtuoso, from a quick glance at the documentation, is that it only works with react projects, and we're trying to maintain support for non react projects, without having to bundle in all of react.

It *seems* like for us it should be possible to calculate column widths since we can just truncate any overflowing text that exceed column width? The dynamic layout function probably just needs to aggregate some statistics on the average length of characters for each column

Open to suggestions though!

wszeborowskimateusz:
Would be available in version 0.3.2

PR #203, opened 2025-05-21
Implement virtual table with tanstack virtual

taehee-kim-sherpas:
https://github.com/HelloCSV/HelloCSV/issues/199

- a lot of a11y enhancement
- implement virtual table with tanstack virtual
- calculate table column width manually
- use portal for sheet tooltip (because of row and cell's tooltip are both `absolute`)


taehee-kim-sherpas:
I create some E2E test using my own test framework (called siheom). It's originally written in clojurescript, It's typescript rewrite.

https://github.com/taehee-kim-sherpas/HelloCSV/tree/feature/siheom-e2e-testing

src/components/Input.tsx
```diff
@@ -18,6 +19,7 @@ interface Props {
   classes?: string;
   clearable?: boolean;
   type?: 'text' | 'number';
+  'aria-label'?: string;
```
wszeborowskimateusz: Maybe we should call it ariaLabel like other props and use it the same way?

src/components/SheetTooltip.tsx
```diff
@@ -53,11 +75,49 @@ export default function SheetTooltip({
...
+      aria-invalid={variant === 'error'}
+      aria-errormessage={variant === 'error' ? tooltipId : undefined}
+      aria-describedby={variant === 'error' ? tooltipId : undefined}
+      onMouseEnter={mouseEnter}
+      onMouseLeave={mouseLeave}
+    >
       {children}
-      {tooltipText && <span className={tooltipClassName}>{tooltipText}</span>}
+      {tooltipText && showTooltip && (
+        <Portal>
```
wszeborowskimateusz: This whole thing doesn't seem to work for me in the example-preact. I don't see the tooltip at all

wszeborowskimateusz: In the example-react it seems to show up, but behaves weirdly


Uploading Screen Recording 2025-05-21 at 9.48.22 AM.mov…

wszeborowskimateusz: Why exactly did we need to make this change?

src/components/SheetTooltip.tsx
```diff
@@ -53,11 +75,49 @@ export default function SheetTooltip({
     withOutline: !!tooltipText,
   });
 
+  const [coords, setCoords] = useState({ left: 0, top: 0, width: 0 });
+  const [showTooltip, setShowTooltip] = useState(false);
+
+  const mouseEnter = (e: MouseEvent<HTMLElement>) => {
```
wszeborowskimateusz: I think this doesn't seem to be working on mobile phone - they don't use mouse 😓

src/i18/types.ts
```diff
@@ -30,3 +25,47 @@ type CreateTranslation<T> = {
...
 export type Translation = Prettify<CreateTranslation<typeof enTranslation>>;
+
+type Join<K, P> = K extends string | number
+  ? P extends string | number
+    ? `${K}.${P}`
+    : never
+  : never;
+
+// Up to 4 levels deep
+type TranslationPaths<T> = T extends string
```
wszeborowskimateusz: Cannot we somehow use recursion here to support any number of nestings?

wszeborowskimateusz: Something like 

```
type TranslationPaths<T> = T extends object
  ? {
      [K in keyof T]: T[K] extends string
        ? K & string
        : T[K] extends object
          ? (K & string) | Join<K & string, TranslationPaths<T[K]>>
          : never;
    }[keyof T]
  : never;
```

src/status/utils.ts
```diff
@@ -30,6 +30,8 @@ function getCsv(rows: SheetRow[]) {
 }
 
 function exportCsv(sheetId: string, rows: SheetRow[]) {
+  if (rows.length === 0) return;
```
wszeborowskimateusz: Why did we make this change? Shouldn't we download file with just headers in such case? Right now it might look like a bug, that the button might not do anything? Especially when all the sheets (in exportAllCsvs) are empty

src/uploader/components/RequirementsList.tsx
```diff
@@ -14,30 +14,34 @@ export default function RequirementsList({ importerRequirements }: Props) {
...
-                >
-                  <div className="text-xs">{requirement.columnLabel}</div>
-                  <div className="text-xs font-light">
-                    <Tooltip
-                      tooltipText={t(`uploader.${groupName}ColumnsTooltip`)}
-                    >
-                      <InformationCircleIcon className="size-5 text-gray-500" />
-                    </Tooltip>
+        .map(([groupName, requirements]) => {
+          const group = groupName === 'required' ? 'required' : 'optional';
```
wszeborowskimateusz: What is the reason behind this change?

src/sheet/components/SheetDataEditorTable.tsx
```diff
@@ -79,7 +130,14 @@ export default function SheetDataEditorTable({
             </th>
 
             {headerGroup.headers.map((header) => (
-              <th key={header.id} className={`z-10 ${headerClass}`}>
+              <th
+                key={header.id}
+                className={`z-10 ${headerClass}`}
+                style={{
+                  width: columnWidthDict[header.column.id],
```
wszeborowskimateusz: How does this work when element doesn't fit the column width? If we e.g. have a long text user won't be able to read it? This is I assume, especially the case when field is read only

src/components/SheetTooltip.tsx
```diff
@@ -53,11 +75,49 @@ export default function SheetTooltip({
...
+      aria-errormessage={variant === 'error' ? tooltipId : undefined}
+      aria-describedby={variant === 'error' ? tooltipId : undefined}
+      onMouseEnter={mouseEnter}
+      onMouseLeave={mouseLeave}
+    >
       {children}
-      {tooltipText && <span className={tooltipClassName}>{tooltipText}</span>}
+      {tooltipText && showTooltip && (
+        <Portal>
+          <span
```
wszeborowskimateusz: It also looks buggy when you click the element - currently it is highlighted - the intention was that it also shows the tooltip then, but now it only shows it on mouse hover

src/sheet/components/SheetDataEditorTable.tsx
```diff
@@ -79,7 +130,14 @@ export default function SheetDataEditorTable({
             </th>
 
             {headerGroup.headers.map((header) => (
-              <th key={header.id} className={`z-10 ${headerClass}`}>
+              <th
+                key={header.id}
+                className={`z-10 ${headerClass}`}
```
wszeborowskimateusz: (comment not necessarily to this particular line)
Screenshot 2025-05-21 at 10 09 53 AM


Seems like those changes might sometimes cause empty space to show up?

src/sheet/components/SheetDataEditorTable.tsx
```diff
@@ -115,11 +173,23 @@ export default function SheetDataEditorTable({
...
+      <tbody
+        className="grid w-full divide-y divide-gray-200"
+        style={{ height: `${rowVirtualizer.getTotalSize()}px` }}
+      >
+        {visibleRows.map(({ row, index, start }) => (
+          <tr
+            key={row.id}
+            data-index={index}
+            ref={(node) => rowVirtualizer.measureElement(node)}
+            className="absolute flex w-full"
```
wszeborowskimateusz: Seems like field input is no longer centered in the cell?

src/components/Input.tsx
taehee-kim-sherpas: I understand it can look weird, but it has historical reason in react ecosystem

> Pitfall
> For historical reasons, aria-* and data-* attributes are written as in HTML with dashes.

https://react.dev/learn/writing-markup-with-jsx#3-camelcase-salls-most-of-the-things

> aria-*: ARIA attributes let you specify the accessibility tree information for this element. See ARIA attributes for a complete reference. In React, all ARIA attribute names are exactly the same as in HTML

https://react.dev/reference/react-dom/components/common#common-props

src/i18/types.ts
taehee-kim-sherpas: Typescript compiler doesn't like infinite recursion. But... i can try to find some workaround

src/status/utils.ts
taehee-kim-sherpas: Now we find the headers from first row. (It's the bug's cause) I will try to find alternative and add more test case

src/uploader/components/RequirementsList.tsx
taehee-kim-sherpas: Now we strictly check transactions key type. And typescript compiler doesn't like dynamic string. So when we create dynamic string key, we have to use string literal union type.

src/sheet/components/SheetDataEditorTable.tsx
taehee-kim-sherpas: Instead of using title, we can use some tooltip. Actually table is not ideal UI for very long text

taehee-kim-sherpas:
Thank you for feedbacks! I'll add the test cases and fix them all.

src/components/SheetTooltip.tsx
taehee-kim-sherpas: fixed! It's scroll issue.

createPortal is needed, because...  the row itself is `absolute` now. previous row's tooltip is overlapped by next row

src/components/SheetTooltip.tsx
taehee-kim-sherpas: fixed We show tooltip on Hover or Focused

src/components/SheetTooltip.tsx
taehee-kim-sherpas: fixed! We show tooltip on Hover or Focused

src/sheet/components/SheetDataEditorTable.tsx
taehee-kim-sherpas: image


 fixed!

src/sheet/components/SheetDataEditorTable.tsx
taehee-kim-sherpas: image

fixed!

src/i18/types.ts
wszeborowskimateusz: And the above thing doesn;t work? I think it was working for me when I tested it?

src/sheet/components/SheetDataEditorTable.tsx
wszeborowskimateusz: I can still see the empty space at the end of the table - this time its shown even on larger screen

https://github.com/user-attachments/assets/23d7340a-559b-44b3-8c64-200efe447bd0

wszeborowskimateusz: And the columns headers seems shifted to the right from the content of the table

src/components/SheetTooltip.tsx
wszeborowskimateusz: This still doesn't look right to me - checkout the video

https://github.com/user-attachments/assets/56a7a81e-11c9-408d-a51d-0983a88101c6

src/sheet/components/SheetDataEditorTable.tsx
wszeborowskimateusz: Yeah, cause right now it seems like long, read only texts would not be fully visible by the user

src/importer/index.tsx
```diff
@@ -270,6 +270,9 @@ function ImporterBody({
         {(mode === 'submit' || mode === 'failed' || mode === 'completed') && (
           <ImportStatus
             mode={mode}
+            headers={
+              columnMappings?.map((mapping) => mapping.sheetColumnId) ?? []
```
wszeborowskimateusz: I don't think this is correct, since thats mappings from all sheets and you don't filter it later anyhow to be sheet specific. Plus there are columns that would not be mapped - like readonly columns, calculated columns, reference columns, non required columns.

I think columns could be fetched from sheet definitions? Checkout sheet/utils.ts::downloadSheetAsCsv - perhaps we could use this function for the final download?

wszeborowskimateusz: We might want to add there the escaping bit though

```
  const CsvContent = [
    finalHeaders.join(','),
    ...rows.map((row) =>
      finalHeaders
        .map((h) => {
          let cell = String(row[h] ?? '');

          cell = cell.replace(/"/g, '""');

          if (/[",\n\r]/.test(cell)) {
            cell = `"${cell}"`;
          }

          return cell;
        })
        .join(',')
    ),
  ].join('\n');
```

src/status/utils.ts
wszeborowskimateusz: As suggested in other comment we should probably use sheet/utils.ts::downloadSheetAsCsv or some other common function in the 2 places - downloading csv on table page and summary page

src/sheet/components/SheetDataEditorTable.tsx
taehee-kim-sherpas: can we show both long text value and error message in one tooltip?

wszeborowskimateusz: But we don't always show the tooltip - we only show it in case of things like required field or error

src/components/SheetTooltip.tsx
wszeborowskimateusz: The tooltips look weird when you scroll the content underneath them

https://github.com/user-attachments/assets/b2a8d7d7-e1ce-4a2a-a91a-d3c9c4a79685

src/importer/index.tsx
wszeborowskimateusz: @emily-fields - perhaps we could reuse the downloadSheetAsCsv method in the next PR, after this one is merged?

emily-fields: Yeah, let's plan to do that after this is merged

src/sheet/components/SheetDataEditorTable.tsx
taehee-kim-sherpas: I create some Heauristic like this

Before (Current main, Browser Default)
image

After (calculate column size according to max length value
image

```ts
        size: 32 // padding
          + Math.max(
            column.label.length, // label size
            ...rowData.map(row => row[column.id].toString().length) // max length value
          ) * 7
          + ('isReadOnly' in column && column.isReadOnly ? 16 : 0), // readonly icon
        maxSize: 250, // around 30 characters
```

src/components/SheetTooltip.tsx
taehee-kim-sherpas: i add intersection observer. How about using Floating UI or something?

taehee-kim-sherpas:
Thank you again for your careful review.

It's 2AM in Korea. I'm going to bed, so if you don't have any problems, you can merge right away.

src/components/SheetTooltip.tsx
wszeborowskimateusz: Well, the tooltips still don't look right - they stay in the same place when we scroll the table

src/sheet/components/SheetDataEditorTable.tsx
```diff
@@ -20,9 +26,24 @@ interface Props {
...
 }
 
+function getColumnWidthDict(columns: { id: string; width: number }[]) {
+  const totalWidth = columns.reduce((acc, column) => {
+    return acc + column.width;
+  }, 0);
+
+  return columns.reduce(
+    (acc, column) => {
+      acc[column.id] = `calc(${column.width / totalWidth} * (100% - 96px))`;
```
wszeborowskimateusz: Where did we got the 96px from?

src/sheet/components/SheetDataEditorTable.tsx
```diff
@@ -63,16 +85,45 @@ export default function SheetDataEditorTable({
...
+    'flex text-sm font-medium whitespace-nowrap text-gray-900 border-b border-gray-300 shrink-0 relative isolate justify-start items-center';
+
+  // The scrollable element for your list
+
+  const rows = table.getRowModel().rows;
+  // The virtualizer
+  const rowVirtualizer = useVirtualizer({
+    count: rows.length,
+    getScrollElement: () => tableContainerRef.current,
+    estimateSize: () => 52.62,
```
wszeborowskimateusz: Where did we got this number from?

src/sheet/components/SheetDataEditorTable.tsx
wszeborowskimateusz: This does seem a buggy - it would be easy to forget about some elements impacting the width

src/components/SheetTooltip.tsx
wszeborowskimateusz: Perhaps doing something like https://github.com/TanStack/virtual/discussions/476 could refrain us from using position:absolute and causing problems with tooltips?

wszeborowskimateusz: And we could keep the old tooltip implementation

src/sheet/components/SheetDataEditorTable.tsx
taehee-kim-sherpas: https://github.com/TanStack/virtual/issues/133#issuecomment-1017778866

Yes...

1. Using floating ui https://floating-ui.com/docs/shift
2. Don't use virtual list (and large data performance issue)
3. reinvent floating-ui ourself

src/sheet/components/SheetDataEditorTable.tsx
taehee-kim-sherpas: It's checkbox column's size

src/sheet/components/SheetDataEditorTable.tsx
taehee-kim-sherpas: It's current row's height

taehee-kim-sherpas: I can extract the hardcoded constant and add some comment.

But I understand that it is buggy and easy to forget to update.

src/sheet/components/SheetDataEditorTable.tsx
wszeborowskimateusz: How about https://github.com/TanStack/virtual/discussions/476?

taehee-kim-sherpas: it looks weird hack, but actually working. I rollbacked the portal implementation

src/utils/hooks.ts
```diff
@@ -60,3 +60,34 @@ export function useLongPress(
     onMouseLeave: clear,
   };
 }
+
+export function useInViewObserver() {
```
wszeborowskimateusz: I think we can remove this since it doesn't seem to be used anymore?

src/sheet/utils.ts
```diff
@@ -152,3 +153,59 @@ export function getEnumLabelDict(sheetDefinitions: SheetDefinition[]) {
...
+function countFullWidth(valueAsString: string | undefined): number {
+  if (valueAsString === undefined) {
+    return 0;
+  }
+
+  return (
+    valueAsString.match(
+      // korean, japanese, chinese, etc.
+      /[ㄱ-ㅎㅏ-ㅣ가-힣ぁ-んァ-ヶ亜-熙一-鿿]/g
+    )?.length ?? 0
```
wszeborowskimateusz: What is the purpose of this function?

src/sheet/components/SheetDataEditor.tsx
```diff
@@ -141,8 +146,12 @@ export default function SheetDataEditor({
         />
       </div>
 
-      <div className="min-h-0 flex-1 overflow-auto">
+      <div
+        className="relative grid min-h-0 overflow-y-auto"
```
wszeborowskimateusz: When we use padding now - do we still need those changes? (replacing flex-1 with relative and grid?)

src/sheet/components/SheetDataEditor.tsx
```diff
@@ -93,8 +93,11 @@ export default function SheetDataEditor({
         header: () => <SheetDataEditorHeader column={column} />,
         sortUndefined: 'last',
         sortingFn: 'auto',
+        size: calculateColumnWidth(column, rowData, enumLabelDict), // readonly icon
```
wszeborowskimateusz: And why exactly do we need to set columns width?

src/sheet/components/SheetDataEditorTable.tsx
wszeborowskimateusz: Cannot we somehow get that dynamically from table definition?

src/sheet/components/SheetDataEditorTable.tsx
wszeborowskimateusz: Yeah, we should definitelly extract it to constant and add some comment. As far as I understand this, this value is not critical and if it is set incorrectly, then we would only got some problems with the scrollbar, right?

src/sheet/components/SheetDataEditorTable.tsx
```diff
@@ -117,11 +184,23 @@ export default function SheetDataEditorTable({
         ))}
       </thead>
 
-      <tbody className="divide-y divide-gray-200">
-        {table.getRowModel().rows.map((row) => (
-          <tr key={row.id}>
+      <tbody
+        className="grid w-full divide-y divide-gray-200"
+        style={{ height: `${rowVirtualizer.getTotalSize()}px`,
+          paddingTop,
```
wszeborowskimateusz: It does seem like some columns are shifted to the right now? Or is it just on my end?

wszeborowskimateusz: This is what I see
Screenshot 2025-05-28 at 7 17 30 AM

src/sheet/components/SheetDataEditorTable.tsx
wszeborowskimateusz: Maybe the best way would be to allow user to resize the columns - https://tanstack.com/table/v8/docs/framework/react/examples/column-sizing ?

wszeborowskimateusz:
I extracted the accessibility changes to
- https://github.com/HelloCSV/HelloCSV/pull/229

wszeborowskimateusz:
We extracted the CSV download part into - https://github.com/HelloCSV/HelloCSV/pull/232

czhu12 (approved):

PR #204, opened 2025-05-21
add more narrative for custom file loaders

czhu12:


docs/common/get-started/custom-file-loaders.mdx
```diff
@@ -0,0 +1,34 @@
...
+}
+```
+
+A `CustomFileLoader` is responsible for parsing a file, of a target `mimeType`, and returning a `csvData` string, which then gets passed
+to the rest of the pipeline.
+
+The `convert` function supports returning a promise for parsing files server-side if needed.
+
+### Usage
+See [customFileLoader](http://localhost:3000/v0.3.1/api-reference/importer-props#customfileloaders) prop.
```
emily-fields: Should this be localhost?

czhu12: Omg, my bad

emily-fields: I think you should remove the localhost part

emily-fields (approved):

PR #218, opened 2025-05-23
feat: use label in enum's reference column

taehee-kim-sherpas:
## Problem

we have a enum field(category-id 카테고리), and a field which reference the enum field(another sheet's category-id)
image
image

but it rendered as value(the actual id), not label in referece column

## Solution

find the reference sheet's enum with enumLabelDict (sheetId -> columId -> enum value -> enum label)

### DIFF
--- a/src/importer/index.tsx
+++ b/src/importer/index.tsx
@@ -24,6 +24,7 @@ import { TranslationProvider, useTranslations } from '../i18';
 import BackToMappingButton from './components/BackToMappingButton';
 import { Uploader } from '../uploader';
 import { convertCsvFile } from '../uploader/utils';
+import { getEnumLabelDict } from '../sheet/utils';
 
 function ImporterBody({
   theme,
@@ -72,24 +73,7 @@ function ImporterBody({
     (sheet) => sheet.id === currentSheetId
   )!;
 
-  const enumLabelDict = Object.fromEntries(
-    sheets.map((sheet) => [
-      sheet.id,
-      Object.fromEntries(
-        sheet.columns
-          .filter((column) => column.type === 'enum')
-          .map((column) => [
-            column.id,
-            Object.fromEntries(
-              column.typeArguments.values.map(({ label, value }) => [
-                value,
-                label,
-              ])
-            ),
-          ])
-      ),
-    ])
-  );
+  const enumLabelDict = getEnumLabelDict(sheets);
 
   const preventUploadOnErrors =
     typeof preventUploadOnValidationErrors === 'function'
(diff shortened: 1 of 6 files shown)


src/sheet/components/SheetDataEditor.tsx
```diff
@@ -27,6 +27,13 @@ interface Props {
   removeRows: (payload: RemoveRowsPayload) => void;
   addEmptyRow: () => void;
   resetState: () => void;
+  enumLabelDict: {
+    [sheetId: string]: {
+      [columnId: string]: {
+        [value: string]: ImporterOutputFieldType;
+      };
+    };
+  };
```
wszeborowskimateusz: This could be extracted to sheet/types.ts and reused in other components

src/sheet/components/SheetDataEditorTable.tsx
```diff
@@ -61,7 +69,7 @@ export default function SheetDataEditorTable({
   }
 
   const headerClass =
-    'bg-hello-csv-muted py-3.5 pr-3 pl-4 text-left text-sm font-semibold text-gray-900 whitespace-nowrap border-y border-gray-300';
+    'bg-hello-csv-muted py-3.5 pr-3 pl-4 text-left text-sm font-semibold text-gray-900 whitespace-nowrap border-y border-gray-300 shrink-0';
```
wszeborowskimateusz: Whats that change for?

src/sheet/components/SheetDataEditorCell.tsx
```diff
@@ -110,8 +118,11 @@ export default function SheetDataEditorCell({
       allData
     );
 
+    const { sheetId, sheetColumnId } = columnDefinition.typeArguments;
+    const labelDict = enumLabelDict[sheetId][sheetColumnId];
+
     const selectOptions = referenceData.map((value) => ({
-      label: value,
+      label: labelDict[value] ?? value,
```
wszeborowskimateusz, line 120: Cannot labelDict be undefined here, since from what I understand you only add columns that are enum columns - so for others enumLabelDict[sheetId][sheetColumnId] would not exist AFAIU

src/importer/index.tsx
```diff
@@ -72,6 +72,25 @@ function ImporterBody({
...
+            column.id,
+            Object.fromEntries(
+              column.typeArguments.values.map(({ label, value }) => [
+                value,
+                label,
+              ])
+            ),
+          ])
+      ),
+    ])
```
wszeborowskimateusz: This whole thing could be extracted to util

src/sheet/components/SheetDataEditorTable.tsx
taehee-kim-sherpas: it is another PR's fix. I removed it.

wszeborowskimateusz (approved):

PR #221, opened 2025-05-26
fix: use enumLabelDict in csv download

taehee-kim-sherpas:
## sheet definition
image

## Before
display column's value

image

## After
display column's label
image


src/sheet/components/SheetDataEditorCell.tsx
```diff
@@ -52,7 +53,9 @@ export default function SheetDataEditorCell({
     columnDefinition.type === 'enum'
       ? (columnDefinition.typeArguments.values.find((e) => e.value === value)
           ?.label ?? value)
-      : value;
+      : columnDefinition.type === 'reference'
```
wszeborowskimateusz: Probably this whole thing (extracting non empty value) could be extracted to a util?

wszeborowskimateusz: Something like `getCellDisplayValue`

src/sheet/utils.ts
```diff
@@ -30,15 +31,26 @@ export function extractReferenceColumnPossibleValues(
...
+        const value = row[column.id];
+        if (column.type === 'enum') {
+          return enumLabelDict[sheetDefinition.id][column.id][value] ?? value;
+        }
+
+        if (column.type === 'reference') {
+          return getLabelDict(column, enumLabelDict)[value] ?? value;
+        }
+        return value;
+      })
```
wszeborowskimateusz: Honestly I am not sure about those changes. I guess my problem is that label is not what is actually stored as value, so if e.g. someone downloads the sheet and tries to reupload it, then it won't work, because the CSV would store the labels and not values

src/sheet/utils.ts
```diff
@@ -30,15 +31,26 @@ export function extractReferenceColumnPossibleValues(
 
 export function downloadSheetAsCsv(
   sheetDefinition: SheetDefinition,
-  data: SheetRow[]
+  data: SheetRow[],
+  enumLabelDict: EnumLabelDict
 ) {
   const headers = sheetDefinition.columns
-    .map((column) => column.id)
+    .map((column) => column.label ?? column.id)
```
wszeborowskimateusz: I am not sure this would work - as noted below.

src/sheet/utils.ts
wszeborowskimateusz: cc: @czhu12

wszeborowskimateusz: @emily-fields

taehee-kim-sherpas: Our users can download it (with label) after upload it back to hello-csv. Can you tell me what's the case?

wszeborowskimateusz: ohh, you are right, I forgot that we have a mapper that checks for enum labels 🤦

src/sheet/utils.ts
wszeborowskimateusz: ok, I made it a config: `csvDownloadMode: 'value' | 'label'`

emily-fields (approved):

PR #232, opened 2025-05-29
Extract shared CSV generation logic

emily-fields:


src/utils/index.ts
```diff
@@ -34,3 +43,67 @@ export function normalizeValue(
...
+        if (column.type === 'enum') {
+          return enumLabelDict[sheetDefinition.id][column.id][value] ?? value;
+        }
+
+        if (column.type === 'reference') {
+          return getLabelDict(column, enumLabelDict)[value] ?? value;
+        }
+        return value;
+      })
+      .join(DOWNLOADED_CSV_SEPARATOR)
```
wszeborowskimateusz, line 94: I think we should escape both headers and rows? Like what if some field contains a string with commas or new lines? This would mess up the csv

src/status/utils.ts
```diff
@@ -1,58 +1,58 @@
...
+  csvDownloadMode: CsvDownloadMode
+) => {
   if (!sheetData.length) return 0;
 
   return sheetData.reduce((totalSize, sheet) => {
-    return totalSize + getCsvSize(sheet.rows);
+    const sheetDefinition = sheetDefinitions.find(
+      (s) => s.id === sheet.sheetId
+    );
+    if (!sheetDefinition) return totalSize;
```
wszeborowskimateusz, line 46: Why do we do it? I think even if sheet is empty we still generate csv file with just headers? Or is this function suppose to count just the data? But in such a case I think it also counts headers 🤔

src/utils/index.ts
```diff
@@ -34,3 +43,86 @@ export function normalizeValue(
       ''
     );
 }
+
+function escapeCsvCell(value: any): string {
+  if (value == null) {
+    return '';
+  }
+
+  let cell = String(value);
```
wszeborowskimateusz, line 52: What happens when value is an integer? Does it work correctly when you download and reupload the file? Or would it become string?

src/status/utils.ts
emily-fields: This function counts the headers and data for multiple sheets if there are multiple sheets. We need this in case the sheet is not found

src/utils/index.ts
emily-fields: I just tested this and I was able to upload the file after downloading it without issues

wszeborowskimateusz (approved):

PR #253, opened 2025-07-22
feat: Add korean translation (locale)

welikeheon:
I'm really pleased with HelloCSV – it's a well-made tool.

I recently integrated it into my internal project and have been using it successfully.
As part of that work, I've added Korean locale support, which I hope will be beneficial to other users.

Cheers,

### DIFF
--- a/src/i18/index.tsx
+++ b/src/i18/index.tsx
@@ -1,6 +1,7 @@
 import { createContext, Fragment } from 'preact';
 import enTranslation from './en/translation.json';
 import frTranslation from './fr/translation.json';
+import koTranslation from './ko/translation.json';
 import ptBRTranslation from './pt-BR/translation.json';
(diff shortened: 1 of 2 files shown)


wszeborowskimateusz:
Hmm, we do now support passing `translationResources`. I am wondering if we want users to use it, instead of integrating more translations into the package itself. We would then need to maintain them, which might not be simple, not knowing korean. Thoughts @czhu12 @emily-fields ?

bertyhell:
@wszeborowskimateusz it might be nice to allow those translation files into the repo, but ndon't bundle them. then users can copy those files to use the languages they need.
and they can be maintained by the community.

otherwise everyone that uses the library needs to reinvent the wheel.


i can deliver the Dutch (nl) translations for this

PR #260, opened 2025-07-31
fix: duplicate react keys in Select component

nayemalam:
## Problem
The Select component generates duplicate React keys when using object values, causing the warning: 
```
Encountered two children with the same key, `[object Object]`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.
```

Here's a reproducible codesandbox as well:
https://codesandbox.io/p/sandbox/rrfwlt

---

This happens specifically in the column mapping step when `MapperOptionValue` objects are used as keys.

## Solution
Updated the key generation logic to create unique string keys:
- For object values: stringified object
- For primitive values: `String(value)` (maintaining existing behavior for strings, numbers, etc.)

This approach ensures unique keys while using meaningful, readable identifiers based on the option labels.

## Reproduction Steps
1. Upload a CSV file with multiple columns
2. Open browser console
3. Click on any select dropdown in the column mapping screen
4. See the duplicate key warning (before fix) or no warning (after fix)

### DIFF
--- a/src/components/Select.tsx
+++ b/src/components/Select.tsx
@@ -1,18 +1,18 @@
 import {
   Combobox,
   ComboboxButton,
+  ComboboxInput,
   ComboboxOption,
   ComboboxOptions,
-  ComboboxInput,
 } from '@headlessui/react';
 import {
+  CheckIcon,
   ChevronUpDownIcon,
   XMarkIcon,
-  CheckIcon,
 } from '@heroicons/react/20/solid';
-import { useTranslations } from '../i18';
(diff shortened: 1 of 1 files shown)


nayemalam:
@wszeborowskimateusz tagging for viz

wszeborowskimateusz:
Nice one. Thanks for contribution 🚀

wszeborowskimateusz (approved):

Issue #264, opened 2025-08-06
Feature Request: Support onFileUpload callback for file metadata


hadedeji10k:
It would be helpful to have more info over the file uploaded. Currently, there is no exposed callback to know:

- When a file is uploaded
- What file was uploaded (its metadata)

wszeborowskimateusz:
Hmm interesting, what is your usecase for it @hadedeji10k ?

hadedeji10k:
@wszeborowskimateusz 
- be able to show user the name, size of the file they uploaded
- tracking the files uploaded by user (saving that to database)

wszeborowskimateusz:
Hmm @hadedeji10k , we currently do expose the uploaded file in the onComplete callback. 

Won't that be enough?

hadedeji10k:
That solves the second use case though, the first use case, will be helpful, when devs need to know the file details prior to submission, but i guess the `onStateChanged` covers that also, since when the file changes, the state changes.. and we can pass the importer state.

wszeborowskimateusz:
Should be resolved in v.0.3.6

PR #265, opened 2025-08-07
add onModeChange callback

rvhoyt:
Adds this optional callback to importer props
`onModeChange?: (previousMode: ImporterMode, newMode: ImporterMode) => void;`
so that component can be better integrated into pages.

### DIFF
--- a/docs/v0.3.5/api-reference/importer-props.mdx
+++ b/docs/v0.3.5/api-reference/importer-props.mdx
@@ -10,7 +10,7 @@ This section details each property available for the Importer component.
 
 **Required:** Yes
 
-**Description:**  
+**Description:**
 Array of sheet definitions for importing data. This property defines the structure of the data to be imported.
 
 ---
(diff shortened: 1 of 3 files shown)


wszeborowskimateusz:
@rvhoyt what is the use case you had in mind? Perhaps we could extend it to the whole state change - like onStateChanged or something like this. I think then it would also incorporate https://github.com/HelloCSV/HelloCSV/issues/264 this request from @hadedeji10k

hadedeji10k:
@wszeborowskimateusz 

I guess `onModeChange` is kind of needed, a case where devs need to know the current status of the import, at entering a specific stage, they might want to show a warning or alert to users.

Merging `onModeChange` to my request to become `onStateChanged` makes sense if it will return the state of the importer, since the state of the import has the mode

rvhoyt:
That's pretty much it, we need to expressly give additional information at each stage of the import and would use this so the rest of our app can update based on what step they're in. Our key purpose just needs to know the mode it's in, any other information to pass could benefit other cases like @hadedeji10k

wszeborowskimateusz:
Ok, closing this one in favor of more general https://github.com/HelloCSV/HelloCSV/pull/267

Issue #287, opened 2025-10-30
Unique validator is case sensitive


rvhoyt:
The unique validator is case sensitive since it internally just uses Set. And since custom validators run a single function per row, there's no way to do a custom one to have the same functionality.
Suppose options are

make new uniqueCaseInsensitive validator,
or add an option to existing unique,
or expand custom validator to have some kind of scope to allow doing it there.

wszeborowskimateusz:
Yeah, I think perhaps we could add a caseSensitive boolean parameter to the UniqueValidator

rvhoyt:
Did a quick run through using Codex because I'm not that familiar with how things are setup with the validators. Seems to work for me. See #288

wszeborowskimateusz:
Done in https://github.com/HelloCSV/HelloCSV/pull/288

wszeborowskimateusz:
@rvhoyt available in v0.4.0

PR #288, opened 2025-11-03
Add case insensitive option for unique validator

rvhoyt:
For Issue:  #287

### DIFF
--- a/src/validators/types.ts
+++ b/src/validators/types.ts
@@ -21,8 +21,15 @@ export type ImporterValidatorType =
   | 'postal_code'
   | 'custom';
 
+export interface UniqueValidatorDefinition
+  extends ImporterValidatorDefinitionBase {
+  validate: 'unique';
+  caseInsensitive?: boolean;
+}
+
 export type ImporterValidatorDefinition =
   | RequiredValidatorDefinition
+  | UniqueValidatorDefinition
   | ImporterValidatorDefinitionBase
   | MultiIncludesValidatorDefinition
   | IncludesValidatorDefinition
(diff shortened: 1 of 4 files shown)


src/validators/types.ts
```diff
@@ -21,8 +21,15 @@ export type ImporterValidatorType =
   | 'postal_code'
   | 'custom';
 
+export interface UniqueValidatorDefinition
+  extends ImporterValidatorDefinitionBase {
+  validate: 'unique';
+  caseInsensitive?: boolean;
```
wszeborowskimateusz, line 27: Maybe it should be 
- `caseSensitive` 
?

This way the default value (case insensitive) would be backward compatible with what we have right now

wszeborowskimateusz: Actually, I am wrong. Current implementation is caseSensitive

wszeborowskimateusz (approved):

src/validators/validator_definitions/unique_validator.ts
```diff
@@ -1,19 +1,30 @@
...
 
 export class UniqueValidator extends Validator {
   seen: Set<ImporterOutputFieldType>;
 
-  constructor(definition: ImporterValidatorDefinitionBase) {
+  definition: UniqueValidatorDefinition;
+
+  constructor(definition: UniqueValidatorDefinition) {
     super(definition);
+    this.definition = definition;
```
wszeborowskimateusz: Is this needed? Doesn't the base Validator class already doing it?

rvhoyt: You're right, happens via super(), removing.

PR #291, opened 2025-11-17
Email validator error message

rvhoyt:
The error message override does not currently work, it does not get passed from the email validator to the regex validator which is internally used.
This passes the definition along from the email validator to the regex one carrying with it the error message


src/validators/validator_definitions/email_validator.ts
```diff
@@ -1,11 +1,12 @@
+import { EmailValidatorDefinition } from '../types';
 import { RegexValidator } from './regex_validator';
 
 export class EmailValidator extends RegexValidator {
-  constructor() {
+  constructor(definition: EmailValidatorDefinition) {
```
wszeborowskimateusz, line 5: I think we could do the same with `PhoneNumberValidator` and `PostalCodeValidator`?

rvhoyt: Added postal code and phone number to operate in same mechanism along w/ tests

wszeborowskimateusz (approved):

